PR #1, opened 2018-10-08
Properly add git submodule

bibikar:
Fixes compilation of native SVM benchmarks.
Apparently simply adding the files wasn't enough and Git doesn't want to add an empty directory, so I just ran 
```
git submodule add https://github.com/CLIUtils/CLI11.git native/CLI11
```


bibikar:
Thanks!

PR #3, opened 2019-04-26
Add daal4py benchmarks and update native, python benchmarks

bibikar:
@oleksandr-pavlyk 

This PR contains both refactoring and adding new benchmarks since the last PR.
- Universally read from .npy files as described in numpy's [NEP-0001](https://github.com/numpy/numpy/blob/067cb067cb17a20422e51da908920a4fbb3ab851/doc/neps/nep-0001-npy-format.rst). note: The current reader `npyfile.h` used in native code does not really understand dtypes and will simply read the entire array into memory, relying on the caller's interpretation for this data.
- Factor out all dataset generation into `make_datasets.py`
- Move sklearn benches to `sklearn/` for clarity
- Add daal4py benches for correlation/cosine distances, linear/ridge regression, kmeans, SVM
- Add native/sklearn benchmarks for logistic regression with L_BFGS solver and RandomForest regression and classification
- Move kmeans.predict benchmarks into kmeans bench files
- Update thread setting to use daal4py API in sklearn/daal4py benches
- Update license headers


sklearn/kmeans.py
```diff
@@ -0,0 +1,77 @@
...
+        return r
+    return st_func
+
+problem_sizes = [(args.size[0], args.size[1])]
+
+X = np.load(args.filex)
+X_init = np.load(args.filei)
+X_mult = np.vstack((X,) * args.data_multiplier)
+
+kmeans = KMeans(n_clusters=10, n_jobs=int(args.num_threads), tol=1e-16,
```
oleksandr-pavlyk: `n_jobs` is not controlling the number of threads used by DAAL, if we are running this in IDP. What does?

bibikar: on line 30, the following code 

```python
num_threads, daal_version = bench.prepare_benchmark(args)
```

runs both `daal4py.daalinit(nthreads=args.num_threads)` and disables finiteness checking for sklearn.

Do we want to remove the n_jobs argument from `KMeans`?

oleksandr-pavlyk: We can keep it, to ensure that upstream KMeans is run in parallel as well.

PR #5, opened 2019-05-16
Add PCA benchmarks

bibikar:
This PR adds PCA benchmarks and NPY output for PCA.

Benchmarks support `svd_solver`s `daal` and `full`. They were verified against scikit-learn PCA using the NPY serialization.

### DIFF
--- a/native/common.hpp
+++ b/native/common.hpp
@@ -163,6 +163,35 @@ void write_table(dm::NumericTablePtr table,
 
 
 /*
+ * Create a new HomogenNumericTable from a submatrix of an existing
+ * HomogenNumericTable.
+ */
+template <typename T>
+dm::NumericTablePtr
+copy_submatrix(dm::NumericTablePtr src,
+               size_t row, size_t col, size_t rows, size_t cols) {
+
+    dm::BlockDescriptor<T> block;
+    src->getBlockOfRows(0, src->getNumberOfRows(), dm::readOnly, block);
+    T *srcdata = block.getBlockPtr();
+
+    dm::NumericTablePtr dest = dm::HomogenNumericTable<T>::create(
+            cols, rows, dm::NumericTable::doAllocate);
+    dest->getBlockOfRows(0, dest->getNumberOfRows(), dm::readWrite, block);
+    T *destdata = block.getBlockPtr();
+
+    for (int i = 0; i < rows; i++) {
+        for (int j = 0; j < cols; j++) {
+            destdata[i*cols + j] = srcdata[(i+col)*cols + j+row];
+        }
+    }
+
+    return dest;
+
+}
+
+
+/*
  * Generate an array of random numbers.
  */
 double *gen_random(size_t n) {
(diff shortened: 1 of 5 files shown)


fschlimb:
Could you also add a daal4py version?

native/common.hpp
```diff
@@ -133,6 +134,69 @@ dm::NumericTablePtr make_table(T *data, size_t rows, size_t cols) {
...
+
+
+/*
+ * Create a new HomogenNumericTable from a submatrix of an existing
+ * HomogenNumericTable.
+ */
+template <typename T>
+dm::NumericTablePtr
+copy_submatrix(dm::NumericTablePtr src,
+               size_t row, size_t col, size_t rows, size_t cols) {
```
oleksandr-pavlyk: Names should be more descriptive: `size_t row` -> `size_t row_start`, `size_t rows` -> `size_t num_rows`, etc.

oleksandr-pavlyk: Or better yet, the input should be `row_start`, `row_finish`, consistent with NumPy slices, then `num_rows = row_finish - row_start`.

native/pca_bench.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+namespace dn = daal::algorithms::normalization;
+
+
+std::pair<da::pca::ResultPtr, dm::NumericTablePtr>
+pca_fit_daal(double *X, size_t rows, size_t cols, size_t n_components) {
+
+    // Find number of components to use in DAAL
+    if (n_components < 1) {
+        n_components = std::min(n_components, rows);
```
oleksandr-pavlyk: Did you mean to write `n_components = std::min(cols, rows)`  instead?

native/pca_bench.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+#include "common.hpp"
+#include "daal.h"
+
+
+namespace dn = daal::algorithms::normalization;
+
+
+std::pair<da::pca::ResultPtr, dm::NumericTablePtr>
+pca_fit_daal(double *X, size_t rows, size_t cols, size_t n_components) {
```
oleksandr-pavlyk: It's good for now, but in the future, it might make sense to make this templated and run benchmarks for doubles as well as floats.

bibikar: Ok. We can basically do this for all benchmarks.

native/common.hpp
bibikar: Fixed. It's now consistent with numpy slices.

native/common.hpp
```diff
@@ -133,6 +134,69 @@ dm::NumericTablePtr make_table(T *data, size_t rows, size_t cols) {
...
+ */
+template <typename T>
+void write_table(dm::NumericTablePtr table,
+                 const std::string descr, const std::string fn) {
+
+    dm::BlockDescriptor<T> block;
+    table->getBlockOfRows(0, table->getNumberOfRows(), dm::readOnly, block);
+    T *data = block.getBlockPtr();
+
+    size_t shape[] = {table->getNumberOfRows(), table->getNumberOfColumns()};
```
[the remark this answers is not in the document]
bibikar: Done

native/pca_bench.cpp
bibikar: Done!

bibikar:
@fschlimb Done!

PR #8, opened 2019-06-13
Use L-BFGS-B Fortran library for native logistic regression benchmark

bibikar:
This PR swaps out the native logistic regression benchmark's solver for the [same one used in SciPy](http://users.iems.northwestern.edu/~nocedal/lbfgsb.html). L-BFGS-B is wrapped in a minimal DAAL `optimization_solver` class and directly set as the solver for DAAL's logistic regression. This brings native performance for this benchmark to match our optimized scikit-learn performance for a sample binary classification problem with 100k samples and 1k features.

Because we must now directly call BLAS and LAPACK functions instead of using DAAL's internal BLAS and LAPACK implementations, we add a dependency on MKL as well.


native/Makefile
```diff
@@ -26,10 +27,19 @@ all: $(addprefix bin/,$(BENCHMARKS))
 bin:
 	mkdir -p bin
 
+bin/log_reg_lbfgs: log_reg_lbfgs_bench.cpp $(FOBJ) | bin
+	$(CXX) $^ $(CXXINCLUDE) $(CXXFLAGS) $(LDFLAGS) -lmkl_rt -lm -lifcore \
```
oleksandr-pavlyk: Linking against `mkl_rt`, are we making sure that MKL is using TBB as the underlying threading layer?

MKL's default is to use OpenMP, as since DAAL's default is to use TBB we end up incurring the cost of runtime of both.

One can either set the threading layer in the benchmark itself, by calling [`mkl_set_threading_layer`](https://software.intel.com/en-us/mkl-developer-reference-c-mkl-set-threading-layer), or use explicit dynamic linking, see [MKL Linking Advisor](https://software.intel.com/en-us/articles/intel-mkl-link-line-advisor).

bibikar: I added a call to `mkl_set_threading_layer` in `native/log_reg_lbfgs_bench.cpp`.

PR #17, opened 2019-11-12
Implement "target time" for benchmarks

bibikar:
On certain benchmarks, the predict method completes orders of magnitude faster than the fit method. Instead of requiring the user to manually select number of repetitions to run, this PR implements aiming for a certain target total time instead. This should help with measurement stability in these benchmarks.

The timing function first takes some "warmup" measurements and determines a good number of inner loops such that the total timing meets the overall time limit and includes about `--goal-outer-loops` outer loops. If the warmup measurement time exceeds the time limit, the timing function just returns the warmup time.

This PR also includes some refactoring of native benchmarks to use common functions for counting classes, printing tables, and timing.

One other change made here is to use `tol=0` in all kmeans benchmarks unless otherwise specified on the command-line. This removes randomness in number of iterations in K-Means fit.

### DIFF
--- a/daal4py/bench.py
+++ b/daal4py/bench.py
@@ -93,12 +93,23 @@ def parse_args(parser, size=None, dtypes=None, loop_types=(),
         else:
             loop_dash = loop_for = ''
 
-        parser.add_argument(f'--{loop_dash}inner-loops', default=3, type=int,
-                            help=f'Number of inner loop iterations {loop_for}'
+        parser.add_argument(f'--{loop_dash}inner-loops', default=100, type=int,
+                            help=f'Maximum inner loop iterations {loop_for}'
                                  f'(we take the mean over inner iterations)')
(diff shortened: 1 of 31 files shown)


daal4py/bench.py
```diff
@@ -212,23 +233,71 @@ def time_mean_min(func, *args, inner_loops=1, outer_loops=1, **kwargs):
...
+    if warmup:
         for _ in range(inner_loops):
+            t0 = timeit.default_timer()
             val = func(*args, **kwargs)
-        t1 = timeit.default_timer()
-
-        times[i] = t1 - t0
+            t1 = timeit.default_timer()
+
+            warmup_time += t1 - t0
```
oleksandr-pavlyk: I would reorder these, to save result of `t1-t0`, like so:

```
last_warmup = t1 - t0
warmup_time += last_warmup
```

bibikar: Done, thanks!

PR #19, opened 2020-01-22
Extension of benchmark parameters and output

Alexsandruss:
Addition of next benchmark parameters:
- Data format (pandas DataFrame/numpy array)
- Data order (column/row major)

Other pull request content:
- JSON output format
- cuML Python API benchmarks
- XGBoost benchmarks (Gradient Boosted Trees) with OMP variables setting
- New data passing method
- Box filter as additional time measurement method
- Benchmarks runner with config parser


bibikar:
In general, looks good! Few comments:
- in arg parsing, --check-finite needs an action and should be named
accordingly
- paramaters -> parameters

sklearn/df_clsf.py
```diff
@@ -3,7 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 import argparse
-from bench import parse_args, time_mean_min, print_header, print_row, size_str
+from bench import parse_args, time_mean_min, print_header, print_row,\
```
bibikar: For these multi-line imports, I would like for the project to follow numpy's practice of using parentheses and indenting the inside. See https://github.com/numpy/numpy/blob/master/numpy/linalg/linalg.py#L21

sklearn/df_clsf.py
```diff
@@ -40,9 +41,11 @@
     except ImportError:
         from sklearn.ensemble import RandomForestClassifier
 
-# Load data
-X = np.load(params.filex.name)
-y = np.load(params.filey.name)
+# Load and convert data
+X = convert_data(np.load(params.filex.name),
+    params.dtype, params.data_order, params.data_format)
```
bibikar: `params.dtype` argument should be indented to match the level of `np.load`. See https://www.python.org/dev/peps/pep-0008/#indentation
I would recommend using `flake8` to check for these things.

Also, because `convert_data` has kwargs, I would prefer for them to be passed as keyword arguments rather than positional arguments.

sklearn/df_clsf.py
```diff
@@ -54,25 +57,49 @@
...
                                      outer_loops=params.predict_outer_loops,
                                      inner_loops=params.predict_inner_loops,
                                      goal_outer_loops=params.predict_goal,
                                      time_limit=params.predict_time_limit,
                                      verbose=params.verbose)
 acc = 100 * accuracy_score(y_pred, y)
-print_row(columns, params, function='df_clsf.predict', time=predict_time,
+
+if params.output_format == "csv":
+    print_header(columns, params)
```
bibikar: Maybe it would be possible to factor out this stuff into something in `bench.py`? Maybe results can be packed into some sort of dictionary, maybe the same dictionary used for the json result, and that could then be printed in the expected csv format. The only reason to separate `print_header` and `print_row` originally was to allow for real-time feedback from the benchmark that it had finished the timing of part of the benchmark. If we abandon that entirely, they can also be changed into one function call. For example, `params.size` could turn into `params.shape` and contain the actual shape of the input, not a string representation.

If it is required to rename e.g. `df_clsf` to `decision_forest_classification`, that should also be done for CSV output.

sklearn/df_clsf.py
```diff
@@ -54,25 +57,49 @@
...
+        "stage": "training",
+        "data_format": params.data_format,
+        "data_type": str(params.dtype),
+        "data_order": params.data_order,
+        "rows": X.shape[0],
+        "columns": X.shape[1],
+        "classes": params.n_classes,
+        "time[s]": fit_time,
+        "accuracy[%]": acc,
+        "algorithm_paramaters": dict(clf.get_params())
```
bibikar: Should probably change `paramaters` -> `parameters` unless there is any compatibility reason.

sklearn/df_clsf.py
```diff
@@ -54,25 +57,49 @@
...
+    import json
+
+    res = {
+        "lib": "sklearn",
+        "algorithm": "decision_forest_classification",
+        "stage": "training",
+        "data_format": params.data_format,
+        "data_type": str(params.dtype),
+        "data_order": params.data_order,
+        "rows": X.shape[0],
```
bibikar: Because this is now using JSON, why not simply say `"rows": X.shape`?

sklearn/bench.py
```diff
@@ -71,11 +68,16 @@ def parse_args(parser, size=None, dtypes=None, loop_types=(),
                         help='Output CSV header')
     parser.add_argument('-v', '--verbose', default=False, action='store_true',
                         help='Output extra debug messages')
-
-    if dtypes is not None:
-        parser.add_argument('-d', '--dtype', default=np.dtype(dtypes[0]),
-                            type=np.dtype, choices=dtypes,
-                            help='Data type to use (see numpy.dtype docs)')
+    parser.add_argument("--data-format", type=str, default="numpy",
```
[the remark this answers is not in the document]
Alexsandruss: Arguments are corrected, single quotes are added

sklearn/bench.py
```diff
@@ -113,6 +115,15 @@ def parse_args(parser, size=None, dtypes=None, loop_types=(),
 
     params = parser.parse_args()
 
+    if params.dtype in ["float64", "double"]:
```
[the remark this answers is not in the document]
Alexsandruss: Data type argument type was changed back to np.dtype

sklearn/bench.py
```diff
@@ -302,4 +313,26 @@ def accuracy_score(y, yp):
     return np.mean(y == yp)
 
 
-sklearn_set_no_input_check()
+def rmse_score(y, yp):
+    return np.sqrt(np.mean((y - yp)**2))
+
+
+def convert_data(data, fptype=np.float64, data_order:str="C", data_format:str="numpy"):
+    if data_order == "F":
```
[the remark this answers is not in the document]
Alexsandruss: Unnecessary error raises were deleted

sklearn/df_clsf.py
Alexsandruss: Import were corrected (at least, in files of this pull request)

make_datasets.py
```diff
@@ -14,8 +14,15 @@ def gen_regression(args):
                            bias=rs.normal(0, 3),
                            random_state=rs)
 
-    np.save(args.filex, X)
```
bibikar: I think I missed this last time, but at the top of this file, there should be the usual copyright header

```python
# Copyright (C) 2018-2020 Intel Corporation
# 
# SPDX-License-Identifier: MIT
```

Part of the code for some of these files dates back to 2017, so that's why I have been adding earlier years as well.

make_datasets.py
```diff
@@ -14,8 +14,15 @@ def gen_regression(args):
...
-    np.save(args.filey, y)
+    if args.test_samples != 0:
+        train_samples = args.samples - args.test_samples
```
bibikar: Some of this code is redundant. If `args.test_samples` is zero, `train_samples` would be the same as `args.samples`, so the train part of the train-test split would still be the exact same. It would be better to put this `train_samples = ...` calculation outside of the if-block, and also move the saving of `args.filex` and `args.filey` out.

Same goes for `gen_classification` and `gen_kmeans`.

make_datasets.py
```diff
@@ -112,6 +133,10 @@ def main():
                              required=True, help='Path to save matrix X')
     regr_parser.add_argument('-y', '--filey', '--fileY', type=str,
                              required=True, help='Path to save vector y')
+    regr_parser.add_argument('-xt', '--filextest', '--fileXtest', type=str,
```
bibikar: Short options must be only one character long. Generally, `-xt` means the same thing as `-x -t`. I would suggest something like `-X` instead of `-xt`. Same goes for `-ts` - could this be `-t`? `-T`? `-s`? `-S`?

sklearn/bench.py
```diff
@@ -298,8 +318,124 @@ def logverbose(msg, verbose):
...
+                params.dtype,
+                params.data_order,
+                params.data_format
+            )
+        # generate data if it's marked and path isn't specified
+        if full_data[element] is None and element in generated_data:
+            full_data[element] = convert_data(np.random.rand(*params.shape),
+                                              params.dtype, params.data_order,
+                                              params.data_format)
+        if full_data[element] is not None and 'y' in element and label_2d and hasattr(full_data[element], 'reshape'):
```
bibikar: Would be useful to have comments before each one of these if-blocks like you did in line 387.

cuml/kmeans.py
```diff
@@ -0,0 +1,99 @@
...
+predict_time, _ = time_mean_min(kmeans.predict, X_test,
+                                outer_loops=params.predict_outer_loops,
+                                inner_loops=params.predict_inner_loops,
+                                goal_outer_loops=params.predict_goal,
+                                time_limit=params.predict_time_limit,
+                                verbose=params.verbose)
+test_inertia = kmeans.inertia_
+
+if params.output_format == 'csv':
+    output_csv(columns, params, functions=['KMeans.fit', 'KMeans.predict'],
```
bibikar: This is a good change from `print_row` and `print_header`, but I think it would be great if there existed one function like `print_result(..., format=...)` which would handle the differences between CSV and JSON formats elsewhere. Preserving the same names of benchmarks is not critical. We could have function be `kmeans` in CSV outputs as well, as long as there is a way to differentiate between benchmarks and `fit`/`predict` routines.

If this is too much for this PR, it could be postponed until later.
Also, was there a reason that the outputs were moved to the very end instead of performing output as soon as timing results are available? I like to be able to see progress when manually running benchmarks, and that was the reason I originally wrote it using separate `print_header` and `print_csv`. If you think that would just make it too messy, that's fine.

runner.py
```diff
@@ -0,0 +1,149 @@
+import argparse
```
bibikar: Same as for `make_datasets.py`, `runner.py` should have a copyright header at the top, but with only 2020.

runner.py
```diff
@@ -0,0 +1,149 @@
...
+            cases[prev_lenght * i + j] += ' {}{} {}'.format(
+                '-' if len(param_name) == 1 else '--',
+                param_name, params[param_name][i])
+    del params[param_name]
+    gen_cases(params)
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--config', metavar='ConfigPath', type=str,
+                    default='config.json')
```
bibikar: Help text would be useful here.

sklearn/bench.py
```diff
@@ -71,16 +78,35 @@ def parse_args(parser, size=None, dtypes=None, loop_types=(),
...
-    if dtypes is not None:
-        parser.add_argument('-d', '--dtype', default=np.dtype(dtypes[0]),
-                            type=np.dtype, choices=dtypes,
-                            help='Data type to use (see numpy.dtype docs)')
+    parser.add_argument('--data-format', type=str, default='numpy',
+                        choices=('numpy', 'pandas', 'cudf'),
+                        help='Data format: numpy, pandas, cudf')
+    parser.add_argument('--data-order', type=str, default='C',
+                        choices=('C', 'F'),
+                        help='Data order: F (column-major) or C (row-major)')
```
bibikar: It would be useful here to include what the default choice is, e.g. with adding something like `(default: %(default)s)` to the help text.

runner.py
Alexsandruss: Header was added

runner.py
Alexsandruss: Added

sklearn/bench.py
Alexsandruss: Added

sklearn/bench.py
Alexsandruss: Comments were added.

make_datasets.py
Alexsandruss: Header was added

make_datasets.py
Alexsandruss: Code was corrected.

make_datasets.py
Alexsandruss: Parser arguments were corrected.

cuml/kmeans.py
```diff
@@ -0,0 +1,99 @@
...
+params = parse_args(parser, prefix='cuml', loop_types=('fit', 'predict'))
+
+# Load and convert generated data
+X_train, X_test, _, _ = load_data(params)
+
+if params.filei is not None:
+    X_init = np.load(params.filei).astype(params.dtype)
+    params.n_clusters = X_init.shape[0]
+else:
+    np.random.seed(params.seed)
```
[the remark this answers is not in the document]
Alexsandruss: Comments were added.
Init algorithm might be added, but will give different centroids for cuML and DAAL. Since threshold is set to 0 by default, it will not affect training performance, but otherwise one of libraries possibly will perform lesser number of iterations.
So, I think that for this pull request this initialization method is enough.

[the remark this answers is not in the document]
bibikar: Ok, sounds good - thanks for investigating. Whenever I run this benchmark, I always import cluster centers from the file, so what you have here is fine for my usage.

cuml/kmeans.py
Alexsandruss: CSV/JSON outputs were merged in one function.

sklearn/df_clsf.py
Alexsandruss: CSV/JSON outputs were merged in one function and moved to bench.py. I think it will be enough for this pull request.

Alexsandruss:
@bibikar, I applied major things/comments for this pull request. Are there other changes that should be in it?

bibikar:
@Alexsandruss Will take a look, thanks!

bibikar:
The makefile also currently have `755` permissions set. If you plan to delete it, that's fine, but if you don't, it should be set back to `644`.

sklearn/df_clsf.py
bibikar: Looks good, thanks! Benchmark files have now regained their simplicity.

Makefile
```diff
@@ -119,23 +119,22 @@ ARGS_SKLEARN_ridge = 	--size "$(REGRESSION_SIZE)"
...
-			--fileY data/two/y-$(LOGREG_SIZE).npy
-ARGS_SKLEARN_logreg5 =	--fileX data/multi/X-$(LOGREG_SIZE).npy \
-			--fileY data/multi/y-$(LOGREG_SIZE).npy
-ARGS_SKLEARN_dfclf2 = 	--fileX data/two/X-$(DFCLF_SIZE).npy \
-			--fileY data/two/y-$(DFCLF_SIZE).npy
-ARGS_SKLEARN_dfclf5 = 	--fileX data/multi/X-$(DFCLF_SIZE).npy \
-			--fileY data/multi/y-$(DFCLF_SIZE).npy
-ARGS_SKLEARN_dfreg = 	--fileX data/reg/X-$(DFREG_SIZE).npy \
-			--fileY data/reg/y-$(DFREG_SIZE).npy
+ARGS_SKLEARN_kmeans = 	--file-X-train data/clustering/kmeans_$(KMEANS_SIZE).npy \
```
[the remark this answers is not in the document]
Alexsandruss: It's unnecessary at least for python part of Makefile which probably will be removed.

runner.py
```diff
@@ -0,0 +1,170 @@
...
+                'Unknown dataset. Only synthetics are supported now')
+        for lib in libs:
+            for i, case in enumerate(cases):
+                command = 'python {}/{}.py --output-format json{} {}'.format(
+                    lib, algorithm, case, paths)
+                print(command)
+                if not args.dummy_run:
+                    r = subprocess.run(
+                        command.split(' '), stdout=subprocess.PIPE,
+                        stderr=stderr_file, encoding='utf-8')
```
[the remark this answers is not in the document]
Alexsandruss: Changed, now benchmarks stderr is redirected into runner stderr in realtime.

runner.py
```diff
@@ -0,0 +1,170 @@
...
+                    log += r.stdout
+
+# add commas to correct JSON output
+while '}\n{' in log:
```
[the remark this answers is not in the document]
Alexsandruss: Addition of results in output JSON was corrected.

runner.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--config', metavar='ConfigPath', type=str,
+                    default='config.json',
+                    help='Path to config with configuration'
+                         'for benchmarks')
+parser.add_argument('--dummy-run', default=False, action='store_true',
+                    help='Run configuration parser and datasets generation'
+                         'without benchmarks running')
```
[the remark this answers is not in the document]
Alexsandruss: CSV output support was added.

sklearn/bench.py
```diff
@@ -298,8 +321,163 @@ def logverbose(msg, verbose):
...
+    if alg_instance is not None:
+        result.update(
+            {'algorithm_parameters': dict(alg_instance.get_params())})
+    return result
+
+
+def print_output(library, algorithm, stages, columns, params, functions,
+                 times, accuracy_type, accuracies, data, alg_instance=None):
+    if params.output_format == 'csv':
+        output_csv(columns, params, functions, times, [None, accuracies[1]])
```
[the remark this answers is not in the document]
Alexsandruss: Training accuracy was added in CSV format.
I think, better unification of code can be available when backward compatibility will not be needed (if it happen).

runner.py
```diff
@@ -0,0 +1,170 @@
...
+            cases[prev_lenght * i + j] += ' {}{} {}'.format(
+                '-' if len(param_name) == 1 else '--',
+                param_name, params[param_name][i])
+    del params[param_name]
+    generate_cases(params)
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--config', metavar='ConfigPath', type=str,
+                    default='config.json',
```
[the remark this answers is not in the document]
Alexsandruss: Corrected, config added.

runner.py
```diff
@@ -0,0 +1,170 @@
...
+    global cases
+    if len(params) == 0:
+        return cases
+    prev_lenght = len(cases)
+    param_name = list(params.keys())[0]
+    n_param_values = len(params[param_name])
+    cases = cases * n_param_values
+    for i in range(n_param_values):
+        for j in range(prev_lenght):
+            cases[prev_lenght * i + j] += ' {}{} {}'.format(
```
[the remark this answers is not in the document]
Alexsandruss: f-strings are applied in all places.

runner.py
```diff
@@ -0,0 +1,170 @@
...
+                        classes, test_samples, features)
+                    paths += ' --file-X-test {} --file-y-test {}'.format(
+                        x_test_file, y_test_file)
+                    command = 'python make_datasets.py -s {} --ts {} -f {} classification -c {} -x {} -y {} --xt {} --yt {}'.format(
+                        int(train_samples)+int(test_samples),
+                        test_samples, features, classes,
+                        x_train_file, y_train_file,
+                        x_test_file, y_test_file)
+                else:
+                    command = 'python make_datasets.py -s {} -f {} classification -c {} -x {} -y {}'.format(
```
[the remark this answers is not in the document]
Alexsandruss: Dataset generation was rewritten, duplication of code was avoid where is possible.

Alexsandruss:
@bibikar, I applied all previous comments, rewrote runner and some other things, added extension for DBSCAN. Benchmarks were tested on different hardware. Can you review changes?

sklearn/ridge.py
```diff
@@ -14,37 +15,47 @@
...
-predict_time, yp = time_mean_min(regr.predict, Xp,
+predict_time, yp = time_mean_min(regr.predict, X_test,
                                  outer_loops=params.predict_outer_loops,
                                  inner_loops=params.predict_inner_loops,
                                  goal_outer_loops=params.predict_goal,
                                  time_limit=params.predict_time_limit,
                                  verbose=params.verbose)
-print_row(columns, params, function='Ridge.predict', time=predict_time)
+
+if params.output_format == 'json':
```
bibikar: Why do we only report accuracies in JSON outputs?

README.md
```diff
@@ -5,9 +5,16 @@ Python*. See benchmark results [here](https://intelpython.github.io/scikit-learn
 
 ## Prerequisites
 - python and scikit-learn to run python versions
+- pandas when using it's DataFrame as input data format
```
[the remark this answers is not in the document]
Alexsandruss: Corrected.

sklearn/ridge.py
Alexsandruss: Added accuracy calculation for CSV output format, you may correct CSV printing to output accuracies if needed. I didn't that because of backward-compatibility

cuml/bench.py
```diff
@@ -0,0 +1,528 @@
...
+    parser.add_argument('-p', '--prefix', default=prefix,
+                        help='Prefix string, for bookkeeping')
+    parser.add_argument('--header', default=False, action='store_true',
+                        help='Output CSV header')
+    parser.add_argument('-v', '--verbose', default=False, action='store_true',
+                        help='Output extra debug messages')
+    parser.add_argument('--data-format', type=str, default='numpy',
+                        choices=('numpy', 'pandas', 'cudf'),
+                        help='Data format: numpy (default), pandas, cudf')
+    parser.add_argument('--data-order', type=str, default='C',
```
PetrovKP: Why this need?

cuml/bench.py
```diff
@@ -0,0 +1,528 @@
...
+    # Secondly, change format of data
+    if data_format == 'numpy':
+        return data
+    elif data_format == 'pandas':
+        import pandas as pd
+
+        if len(data.shape) == 1:
+            return pd.Series(data)
+        else:
+            return pd.DataFrame(data)
```
PetrovKP: For order format we can different type DataFrame. Do we need to support this?

cuml/bench.py
```diff
@@ -0,0 +1,528 @@
...
+    elif data_format == 'cudf':
+        import cudf
+        import pandas as pd
+
+        return cudf.DataFrame.from_pandas(pd.DataFrame(data))
```
PetrovKP: I think it will be correct if we first create a cupy array, and then create it to the cuda data frame.

cuml/dbscan.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (C) 2020 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import argparse
+from bench import parse_args, time_mean_min, load_data, print_output
+from cuml import DBSCAN
+
+parser = argparse.ArgumentParser(description='scikit-learn DBSCAN benchmark')
```
PetrovKP: cuml?

cuml/knn_clsf.py
```diff
@@ -0,0 +1,52 @@
...
+parser.add_argument('--metric', type=str, default='euclidean',
+                    help='Distance metric to use')
+params = parse_args(parser)
+
+# Load generated data
+X_train, X_test, y_train, y_test = load_data(params)
+params.n_classes = y_train[y_train.columns[0]].nunique()
+
+# Create classification object
+knn_clsf = KNeighborsClassifier(n_neighbors=params.n_neighbors,
```
PetrovKP: cuml KNN is not support double type

cuml/linear.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+import argparse
+from bench import (
+    parse_args, time_mean_min, load_data, print_output, rmse_score
+)
+from cuml import LinearRegression
+
+parser = argparse.ArgumentParser(description='scikit-learn linear regression '
```
PetrovKP: cuml?

cuml/log_reg.py
```diff
@@ -0,0 +1,73 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+import argparse
+from bench import (
+    parse_args, time_mean_min, load_data, print_output, accuracy_score
+)
+from cuml import LogisticRegression
+
+parser = argparse.ArgumentParser(description='scikit-learn logistic '
```
PetrovKP: cuml?

daal4py/bench.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2017-2019 Intel Corporation
+# Copyright (C) 2017-2020 Intel Corporation
 #
```
PetrovKP: Are these copy-paste files in three benchmarks: cuml, daal4py, sklearn? What for? Can't put it in one?

cuml/log_reg.py
Alexsandruss: Corrected

cuml/linear.py
Alexsandruss: Corrected

cuml/knn_clsf.py
Alexsandruss: Corrected

cuml/dbscan.py
Alexsandruss: Corrected

cuml/bench.py
```diff
@@ -0,0 +1,528 @@
...
+
+
+import argparse
+import numpy as np
+import sklearn
+import timeit
+import json
+
+
+def sklearn_disable_finiteness_check():
```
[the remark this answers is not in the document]
Alexsandruss: One bench.py file for all libs.

daal4py/bench.py
bibikar: Originally, this set of benchmarks only had sklearn and native implementations, so this wasn't a problem. But now, we might want to change the directory structure to be proper Python module stuff and access benchmarks using something like `python -m sklearn_bench.sklearn.ridge [ARGS...]`. Then we can use relative imports for this common file. The top-level runner and dataset creation file could also be included in the top-level module.
Switching to using this format isn't a big deal, but can be done in another PR.

sklearn/ridge.py
bibikar: There is no requirement for 1:1 compatibility of CSV formats here. My main requirement is that we should not entirely replace CSV outputs with JSON. If we can write less code by changing CSV outputs slightly, then I think it's a good idea. But not a problem if you need this merged soon.

cuml/bench.py
```diff
@@ -0,0 +1,528 @@
...
+    if verbose:
+        print('@', msg)
+
+
+def convert_to_numpy(data):
+    '''
+    Convert input data to numpy array
+    '''
+    if 'cudf' in str(type(data)):
+        data = data.to_pandas().values
```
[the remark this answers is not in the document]
Alexsandruss: cuDF documentation hasn't this function, cudf.DataFrame too.

Alexsandruss:
@bibikar, I added box filter option for time measurements and XGBoost benchmark. Please, review it and other parts if needed. I think, no more additions are expected for this PR. So, it's to time to finish it.

bibikar:
I tried to run `python runner.py --config config_example.json` and got the following.

```
(idp2020.0) sbibikar@ansatlin13 /localdisk/work/sbibikar/repos/scikit-learn_bench$ python runner.py --config config_example.json
Traceback (most recent call last):
  File "runner.py", line 163, in <module>
    'channel': pkg_info[3]
IndexError: list index out of range
```

It appears that `conda list`'s output sometimes doesn't include the channel:

```
# packages in environment at /localdisk/work/sbibikar/miniconda3/envs/idp2020.0:
#
# Name                    Version                   Build  Channel
absl-py                   0.9.0                    py37_0
asn1crypto                0.24.0                   py37_3    intel
astor                     0.8.0                    py37_0
backcall                  0.1.0                    py37_0
bzip2                     1.0.8                         0    intel
c-ares                    1.15.0            h7b6447c_1001
certifi                   2019.9.11                py37_0
cffi                      1.12.3                   py37_2    intel
chardet                   3.0.4                    py37_3    intel
conda-package-handling    1.4.0                    py37_0    intel
cryptography              2.7                      py37_0    intel
cycler                    0.10.0                   py37_7    intel
cython                    0.29.13          py37ha68da19_0    intel
daal                      2020.0                intel_166    intel
daal4py                   2020.0           py37ha68da19_5    intel
decorator                 4.4.1                      py_0
freetype                  2.10.1                        1    intel
funcsigs                  1.0.2                    py37_7    intel
gast                      0.3.3                      py_0
google-pasta              0.1.8                      py_0
grpcio                    1.27.2           py37hf8bcb03_0
h5py                      2.8.0            py37h989c5e5_3
```

It may be cleaner to also run `conda list --json`, parse its output, and grab what you need (or just grab everything).

runner.py
```diff
@@ -0,0 +1,195 @@
...
+                          + f'--arch {hostname} --header --output-format ' \
+                          + f'{args.output_format}{case} {paths}'
+                while '  ' in command:
+                    command = command.replace('  ', ' ')
+                verbose_print(command)
+                if not args.dummy_run:
+                    stdout, stderr = read_output_from_command(command)
+                    stderr = filter_stderr(stderr)
+                    if args.output_format == 'json':
+                        json_result['results'] += json.loads(f'[{stdout}]')
```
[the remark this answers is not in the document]
Alexsandruss: I changed it to `json_result['results'].extend(json.loads(stdout))`

sklearn/distances.py
```diff
@@ -28,4 +36,18 @@
                             goal_outer_loops=params.goal,
                             time_limit=params.time_limit,
                             verbose=params.verbose)
-    print_row(columns, params, function=metric.capitalize(), time=time)
+    if params.output_format == 'csv':
```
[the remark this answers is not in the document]
Alexsandruss: Since distances benchmark had flow that differs from others, `print_output` wasn't suitable there, but I changed benchmark to work with one metric and now this functions is added.

sklearn/ridge.py
Alexsandruss: I removed `if params.output_format == 'json'`

cuml/bench.py
```diff
@@ -103,6 +103,25 @@ def parse_args(parser, size=None, loop_types=(),
     parser.add_argument('--output-format', type=str, default='csv',
                         choices=('csv', 'json'),
                         help='Output format: csv (default) or json')
+    parser.add_argument('--time-method', type=str, default='mean_min',
+                        choices=('box_filter', 'mean_min'),
+                        help='Method used for time mesurements')
+    parser.add_argument('--n-meas', type=int, default=100,
```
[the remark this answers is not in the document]
Alexsandruss: Arg name is changed to `box-filter-measurements`.

config_example.json
```diff
@@ -0,0 +1,105 @@
...
+        "data-format": ["pandas"],
+        "data-order": ["F"],
+        "dtype": ["float64"]
+    },
+    "cases": [
+        {
+            "algorithm": "distances",
+            "dataset": [
+                {
+                    "training": "synth_clsf_1000_15000_2"
```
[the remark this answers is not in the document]
Alexsandruss: I changed synthetic and others dataset configs to cleaner view.

cuml/bench.py
```diff
@@ -227,6 +217,44 @@ def prepare_daal(num_threads=-1):
     return num_threads, daal_version
 
 
+def measure_function_time(func, *args, params, **kwargs):
```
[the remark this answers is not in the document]
Alexsandruss: No exact need for changing now. Many algorithms don't have verbose mode, but I agree it's good to have.

Alexsandruss:
Thanks for `conda list --json`, I didn't now about it.
OMP_NUM_THREADS are needed in XGBoost and currently sets only for this lib. Without it XGBoost may have lower performance then HT is on.
I looked at all conversations at this pull request, it looks like all comments are applied. Am I right?

Alexsandruss:
@bibikar , are there any stoppers/comments for this pull request?

bibikar:
@Alexsandruss, I will test it again and see if it works for my needs, but recent changes look good :+1:

bibikar:
Have you added functionality to change the number of threads the benchmark can use from the runner? It will be required to execute sklearn benchmarks using `--num-threads` and also provide `OMP_NUM_THREADS` environment variable to sklearn runs.

`getFPType = import_fptype_getter()` seems somewhat unnecessary. You could just put that code at the top-level of `bench.py` and then do `from bench import getFPType` in each benchmark and not have to call the function from each benchmark.

Alexsandruss:
Yes, number of threads is supported. Add `"num-threads": [1]` to common params or single case params in passed config to run benchmarks single-threaded.

Alexsandruss:
Benchmarks with usage of `getFPType` are corrected.

bibikar:
It turns out that using column-major order (which is the default in `config_example.json`) causes sklearn's `RandomForestClassifier` to not call DAAL. Is there any reason that column-major order is the default in the example?

Alexsandruss:
RandomForest isn't patched because of algorithm difference between sklearn and DAAL but implemented as sklearn class. It's a bit confusing, but without direct import (`from daal4py.sklearn.ensemble import RandomForestClassifier`) DAAL isn't used. Argument `--use-sklearn-class` controls which class is used.

Alexsandruss:
Pandas DataFrame with column-major order is chosen for default in config example since it's widely used as input for ML algorithms.

Alexsandruss:
@bibikar, are you still expecting comment from Oleksandr? Can pull request be merged without it?

bibikar:
@Alexsandruss I will ping him and see if he has any comments. Otherwise, it looks ready to me

bibikar:
Merged, thanks!

PR #23, opened 2020-05-17
adding parameters for device context and patching of Scikit-Learn

Alexander-Makaryev:
This is initial suggestion in purpose to start the discussion.
Parameters for device context and Scikit-Learn patching are added.
Some benchmarks are changed in purpose to use new parameters.


Alexander-Makaryev:
Example of the config file:
```
{
    "common": {
        "lib": ["sklearn"],
        "data-format": ["numpy"],
        "data-order": ["C"],
        "device": ["None", "host", "cpu", "gpu"],
        "patch_sklearn": ["False", "True"],
        "dtype": ["float64"]
    },
    "cases": [
                {
            "algorithm": "kmeans",
            "dataset": [
                {
                    "source": "synthetic",
                    "type": "kmeans",
                    "n_clusters": 10,
                    "n_features": 50,
                    "training": {
                        "n_samples": 1000000
                    }
                }
            ],
            "n-clusters": [10]
        },
        {
            "algorithm": "dbscan",
            "dataset": [
                {
                    "source": "synthetic",
                    "type": "blobs",
                    "n_clusters": 10,
                    "n_features": 50,
                    "training": {
                        "n_samples": 10000
                    }
                }
            ],
            "min-samples": [5000],
            "eps": [1]
        },
        {
            "algorithm": "linear",
            "dataset": [
                {
                    "source": "synthetic",
                    "type": "regression",
                    "n_features": 50,
                    "training": {
                        "n_samples": 1000000
                    }
                }
            ]
        },
        {
            "algorithm": "log_reg",
            "solver":["lbfgs", "newton-cg"],
            "dataset": [
                {
                    "source": "synthetic",
                    "type": "classification",
                    "n_classes": 2,
                    "n_features": 100,
                    "training": {
                        "n_samples": 100000
                    }
                },
                {
                    "source": "synthetic",
                    "type": "classification",
                    "n_classes": 5,
                    "n_features": 100,
                    "training": {
                        "n_samples": 100000
                    }
                }
            ]
        }
    ]
}
```

sklearn/bench.py
```diff
@@ -153,6 +153,12 @@ def parse_args(parser, size=None, loop_types=(),
                         help='Seed to pass as random_state')
     parser.add_argument('--dataset-name', type=str, default=None,
                         help='Dataset name')
+    parser.add_argument('--device', type=str, default='None',
```
napetrov, line 156: I guess this params should be enabled in higher level scripts that will execute them - but i don't see this in make? 

Another question - are we going to pass single value for bench.py or this should be tuple so bench.py will iterate via them?

Alexander-Makaryev: At this moment I used only runner.py script. Possible that additional changes should be done in make or somewhere else. Thanks.

Devices are provided as a list in config ``"device": ["None", "host", "cpu", "gpu"],``. See example above. Benchmarks are executing with all devices in the list.

sklearn/bench.py
```diff
@@ -607,3 +613,36 @@ def import_fptype_getter():
...
+def patch_sklearn():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--patch_sklearn', type=str, default='None',
+                        choices=('None', 'True', 'False'),
+                        help='True for patch, False for unpatch, "None" to leave as is.')
+    args, _ = parser.parse_known_args()
+
+    if args.patch_sklearn is not None and args.patch_sklearn != 'None':
+        from daal4py.sklearn import patch_sklearn, unpatch_sklearn
+        if args.patch_sklearn == "True":
```
PetrovKP, line 627: Why are "True", "False" string, not boolean?

PetrovKP:
> Example of the config file:
> 
> ```
> {
>     "common": {
>         "lib": ["sklearn"],
>         "data-format": ["numpy"],
>         "data-order": ["C"],
>         "device": ["None", "host", "cpu", "gpu"],
>         "patch_sklearn": ["False", "True"],
>         "dtype": ["float64"]
>     },
>     "cases": [
>                 {
>             "algorithm": "kmeans",
>             "dataset": [
>                 {
>                     "source": "synthetic",
>                     "type": "kmeans",
>                     "n_clusters": 10,
>                     "n_features": 50,
>                     "training": {
>                         "n_samples": 1000000
>                     }
>                 }
>             ],
>             "n-clusters": [10]
>         },
>         {
>             "algorithm": "dbscan",
>             "dataset": [
>                 {
>                     "source": "synthetic",
>                     "type": "blobs",
>                     "n_clusters": 10,
>                     "n_features": 50,
>                     "training": {
>                         "n_samples": 10000
>                     }
>                 }
>             ],
>             "min-samples": [5000],
>             "eps": [1]
>         },
>         {
>             "algorithm": "linear",
>             "dataset": [
>                 {
>                     "source": "synthetic",
>                     "type": "regression",
>                     "n_features": 50,
>                     "training": {
>                         "n_samples": 1000000
>                     }
>                 }
>             ]
>         },
>         {
>             "algorithm": "log_reg",
>             "solver":["lbfgs", "newton-cg"],
>             "dataset": [
>                 {
>                     "source": "synthetic",
>                     "type": "classification",
>                     "n_classes": 2,
>                     "n_features": 100,
>                     "training": {
>                         "n_samples": 100000
>                     }
>                 },
>                 {
>                     "source": "synthetic",
>                     "type": "classification",
>                     "n_classes": 5,
>                     "n_features": 100,
>                     "training": {
>                         "n_samples": 100000
>                     }
>                 }
>             ]
>         }
>     ]
> }
> ```

Why doesn't the config add to this repository?

sklearn/bench.py
Alexander-Makaryev: One more option is "None" - to leave patching state "as is". think it can be useful for back compatibility. Possible we should change it to boolean.

Alexander-Makaryev:
@PetrovKP At this moment we have only one example in the repository. I suppose to create a directory with various configs. Please comment what do you think about it.

sklearn/bench.py
PetrovKP: What will be broken in this case?

PetrovKP:
> @PetrovKP At this moment we have only one example in the repository. I suppose to create a directory with various configs. Please comment what do you think about it.

We wanted to store all the configs here. This is correct and convenient. I don’t know why we still haven’t moved the configs ...

Alexsandruss:
Implemented in https://github.com/IntelPython/scikit-learn_bench/pull/133

PR #26, opened 2020-05-29
PCA components number will be printed

masdevas:


daal4py/pca.py
```diff
@@ -137,7 +137,7 @@ def test_transform(Xp, pca_result, eigenvalues, eigenvectors):
              params=params, functions=['PCA.fit', 'PCA.transform'],
              times=[fit_time, transform_time], accuracy_type=None,
              accuracies=[None, None], data=[X_train, X_test],
-             alg_params={'svd_solver': params.svd_solver})
+             alg_params={'svd_solver': params.svd_solver, 'n_components': params.n_components})
```
Alexsandruss, line 140: More than 80 symbols in line, please, add line break and use flake8 to check code formatting correctness:
`pip install flake8`
`flake8 daal4py/pca.py`

masdevas: Thank you for formatting request!

PR #27, opened 2020-05-30
Extend datasets for public and inner using

PetrovKP:
1. Add support loader datasets
2. Extend datasets for gisette, a9a, ijcnn and skin-segmentation for SVM algorithm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+# Logs
+*.log
+
+# Release and work directories
+__pycache__*
+__work*
+
+# Visual Studio related files, e.g., ".vscode"
+.vs*
+
+# Datasets
+dataset
+*.csv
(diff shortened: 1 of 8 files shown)


PetrovKP:
@SmirnovEgorRu , @PivovarA , @Alexsandruss  What do you think about this format for receiving datasets?

Alexsandruss:
There must be an option to load dataset directly without saving to csv.

PetrovKP:
@Alexsandruss 
> There must be an option to load dataset directly without saving to csv.

What is it for? When downloading, the original datasets are stored in cache folders.

Alexsandruss:
> > There must be an option to load dataset directly without saving to csv.
> 
> What is it for? When downloading, the original datasets are stored in cache folders.

Some datasets are too heavy to process them, save to csv and read from same csv. Option to load directly (without or with csv saving) is needed to reduce benchmarks running time.

PetrovKP:
Implemented here https://github.com/IntelPython/scikit-learn_bench/pull/44

PR #28, opened 2020-05-30
Fix svm using for sklearn and cuml

PetrovKP:


Alexsandruss:
I have only two comments:
1) Did you test both of scripts with default algorithm parameters?
2) Can you add these changes for daal4py? It looks like there are no significant troubles with that

PetrovKP:
1. Sure, I running for default parameters
2. Thanks, I added

PR #31, opened 2020-07-21
Fix cuml logs for v0.14

PetrovKP:


Alexsandruss:
This fix doesn't work. cuML docs don't provide understandable way to set logger level

Alexsandruss:
Will be fixed by https://github.com/IntelPython/scikit-learn_bench/pull/37

PR #35, opened 2020-08-18
Update of parameters for sklearn & cuml benchmarks

PetrovKP:


daal4py/svm.py
```diff
@@ -210,12 +210,8 @@ def test_fit(X, y, params):
     fptype = getFPType(X)
     kf = daal_kernel(params.kernel, fptype, gamma=params.gamma)
 
-    if params.n_classes == 2:
-        y[y == 0] = -1
-    else:
-        y[y == -1] = 0
-
     svm_train = svm_training(
+            method='thuner',
```
masdevas, line 214: Is it an errata?

PetrovKP: yes, thanks

sklearn/elasticnet.py
```diff
@@ -0,0 +1,51 @@
...
+                        tol=params.tol, max_iter=params.maxiter, copy_X=False)
+
+columns = ('batch', 'arch', 'prefix', 'function', 'threads', 'dtype', 'size',
+           'time')
+
+# Time fit
+fit_time, _ = measure_function_time(regr.fit, X_train, y_train, params=params)
+
+# Time predict
+predict_time, pred_train = measure_function_time(regr.predict, X_train, params=params)
```
owerbat, line 40: Are you sure that we should measure prediction with training dataset? As I see, in other algorithms we use testing dataset for it. (The same question for Lasso and SVM)

daal4py/svm.py
```diff
@@ -324,8 +319,6 @@ def main():
     params.n_classes = np.unique(y_train).size
 
     # This is necessary for daal
```
owerbat, line 321: This comment should be removed, I suppose

sklearn/elasticnet.py
PetrovKP: I would like to measure predictions everywhere on the training set. this is more data, but take into account the test dataset to compare the accuracy. But old measurements and benchmarks cannot be updated - since we are already targeting these cases. But algorithms Lasso, Elastic-Net, SVM - new.

owerbat (approved):
👍

PR #37, opened 2020-10-01
kNN bench, configs and runner update

Alexsandruss:
- Separate training and prediction in kNN benchmarks, add unsupervised support
- Add configs for sklearn, cuML and XGBoost
- Control daal4py sklearn patching by FORCE_DAAL4PY_SKLEARN env variable
- Support of multiple configs in runner


cuml/bench.py
```diff
@@ -8,6 +8,18 @@
 import sklearn
 import timeit
 import json
+import os
+import sys
+
+
+if 'FORCE_DAAL4PY_SKLEARN' in os.environ.keys():
```
PetrovKP, line 15: Add it to Readme

PetrovKP, line 15: Similar to sklearn patch
```suggestion
if os.environ.get('FORCE_DAAL4PY_SKLEARN', False) in [True, '1', 'y', 'yes', 'Y', 'YES', 'Yes', 'true', 'True', 'TRUE']:
```

cuml/bench.py
```diff
@@ -8,6 +8,18 @@
 import sklearn
```
PetrovKP, line 8: do we really need to copy this file to cuml/sklearn/xgboost - given that it is the same everywhere?

configs/cuml_config.json
```diff
@@ -0,0 +1,486 @@
+{
```
PetrovKP, line 1: I would put it in folders:
- sklearn
- xgboost
- cuml
...

cuml/bench.py
Alexsandruss: Not in scope of this PR

configs/cuml_config.json
Alexsandruss: Not in scope of this PR

cuml/bench.py
Alexsandruss: Added

Alexsandruss: Added

PR #38, opened 2020-10-09
Report generator

Alexsandruss:
Excel table generator from json outputs.


PetrovKP (changes requested):
all scripts with the generator move to a separate folder. in this folder, create README.md - how to use.

Alexsandruss:
> all scripts with the generator move to a separate folder. in this folder, create README.md - how to use.

But there is only one script file and one config file, do they need separate folder?

PetrovKP:
> > all scripts with the generator move to a separate folder. in this folder, create README.md - how to use.
> 
> But there is only one script file and one config file, do they need separate folder?

@Alexsandruss , Yes, all auxiliary tools are best placed in a separate folder. in addition, there will be a separate start-up instructions.

PetrovKP (approved):

PR #40, opened 2020-11-13
In-place predict for higgs

ShvetsKS:


ShvetsKS:
@Alexsandruss @PetrovKP could you take a look at updated xgboost benchmark?

xgboost/gbt.py
```diff
@@ -67,6 +67,8 @@ def convert_xgb_predictions(y_pred, objective):
                     help='Control a balance of positive and negative weights')
 parser.add_argument('--count-dmatrix', default=False, action='store_true',
                     help='Count DMatrix creation in time measurements')
+parser.add_argument('--inplace-predict', default=False, action='store_true',
+                    help='Count DMatrix creation in time measurements')
```
Alexsandruss, line 71: Copy-pasted help from previous argument, write own

configs/xgb_cpu_config.json
```diff
@@ -131,7 +131,8 @@
             "n-estimators": [1000],
             "objective": ["binary:logistic"],
             "tree-method": ["hist"],
-            "enable-experimental-json-serialization": ["False"]
+            "enable-experimental-json-serialization": ["False"],
+            "inplace-predict": [""]
```
Alexsandruss, line 135: Are same changes needed for XGB GPU config?

xgboost/gbt.py
```diff
@@ -135,9 +137,13 @@ def fit():
...
-    def predict():
-        dtest = xgb.DMatrix(X_test, y_test)
-        return booster.predict(dtest)
+    if params.inplace_predict == False:
+        def predict():
+            dtest = xgb.DMatrix(X_test, y_test)
+            return booster.predict(dtest)
+    else:
+        def predict():
+            return booster.inplace_predict(np.ascontiguousarray(X_test.values, dtype=np.float32))
```
Alexsandruss, line 146: Why np.ascontiguousarray?

Alexsandruss (changes requested):

xgboost/gbt.py
ShvetsKS: inplace supports only numpy c_contigous array

configs/xgb_cpu_config.json
ShvetsKS: done

xgboost/gbt.py
ShvetsKS: done

xgboost/gbt.py
ShvetsKS: I hope Sklearn API will support this method soon for pandas and numpy.

Alexsandruss: And float32 only?

ShvetsKS: If data is `float64` same conversion will be performed inside `inplace_predict`
So as a conversion is included to measurement I don't think that there is any problem

Alexsandruss (approved):

PR #43, opened 2020-12-11
Update README.md

PetrovKP:


PetrovKP:
fixed here: https://github.com/IntelPython/scikit-learn_bench/pull/44

PR #49, opened 2021-01-11
Small extensions & add new dataset

PetrovKP:
1. Returned support of hardware and software for the logging.
2. Extend cuml config for `skin_segmentation` workload
3. Add loader for `codrnanorm` dataset

### DIFF
--- a/datasets/loader.py
+++ b/datasets/loader.py
@@ -37,7 +37,7 @@ def a9a(dataset_dir=None):
     a9a X train dataset (39073, 123)
     a9a y train dataset (39073, 1)
     a9a X test dataset  (9769,  123)
-    a9a y train dataset (9769,  1)
+    a9a y test dataset  (9769,  1)
     """
     dataset_name = 'a9a'
     os.makedirs(dataset_dir, exist_ok=True)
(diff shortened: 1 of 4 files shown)


datasets/loader.py
```diff
@@ -313,6 +313,31 @@ def covertype(dataset_dir=None):
     return True
 
 
+def codrnanorm(dataset_dir=None):
+    """
+    """
```
Alexsandruss, line 318: Empty comment. Is this place for dataset description?

runner.py
```diff
@@ -106,18 +110,17 @@ def generate_cases(params):
...
                     train_data = dataset["training"]
-                    test_data = dataset["testing"]
-
                     file_train_data_x = train_data["x"]
-                    file_train_data_y = train_data["y"]
-                    file_test_data_x = test_data["x"]
-                    file_test_data_y = test_data["y"]
                     paths = f'--file-X-train {file_train_data_x}'
                     if 'y' in dataset['training'].keys():
+                        file_train_data_y = train_data["y"]
```
Alexsandruss, line 116: `file_train_data_y` is used on line 130 and now might be unassigned.
I guess, need to change line 130 to `if not utils.is_exists_files([file_train_data_x]):`

datasets/loader.py
PetrovKP: Apply

runner.py
PetrovKP: Make sense, thanks! Applied

Alexsandruss (approved):

PR #57, opened 2021-03-17
Add device context parameter

vlad-nazarov:


bench.py
```diff
@@ -175,7 +175,11 @@ def parse_args(parser, size=None, loop_types=(),
                         help='Dataset name')
     parser.add_argument('--no-intel-optimized', default=False, action='store_true',
                         help='Use no intel optimized version. '
-                             'Now avalible for scikit-learn benchmarks'),
+                             'Now avalible for scikit-learn benchmarks')
+    parser.add_argument('--device', default=None, type=str,
```
PetrovKP, line 179:
```suggestion
    parser.add_argument('--device', default="host", type=str,
```

configs/skl_with_context_config.json
```diff
@@ -0,0 +1,77 @@
+{
+    "common": {
+        "lib": ["sklearn"],
+        "data-format": ["pandas"],
+        "data-order": ["F"],
+        "dtype": ["float64"],
+        "device": ["host", "cpu", "gpu"]
```
PetrovKP, line 7: What happens if I run this config on a machine without a GPU driver?

sklearn_bench/dbscan.py
```diff
@@ -19,38 +19,41 @@
...
+
+    # Load generated data
+    X, _, _, _ = bench.load_data(params, add_dtype=True)
+
+    # Create our clustering object
+    dbscan = DBSCAN(eps=params.eps, n_jobs=params.n_jobs,
+                    min_samples=params.min_samples, metric='euclidean',
+                    algorithm='auto')
+
+    # N.B. algorithm='auto' will select DAAL's brute force method when running
```
PetrovKP, line 35:
```suggestion
    # N.B. algorithm='auto' will select oneAPI Data Analytics Library (oneDAL) brute force method when running
```

PetrovKP (commented):
overall good, but need to check the performance of the cpu (so that your changes do not bring an overhead). come to me

bench.py
michael-smirnov: My understanding is that `None` is used to run without context. Other values specify device type for a context

bench.py
```diff
@@ -175,7 +175,11 @@ def parse_args(parser, size=None, loop_types=(),
...
+                        choices=("host", "cpu", "gpu"),
```
michael-smirnov, line 180: `None` shall be also included?

bench.py
```diff
@@ -197,6 +201,8 @@ def parse_args(parser, size=None, loop_types=(),
         except ImportError:
             print('Failed to import daal4py.sklearn.patch_sklearn.'
                   'Use stock version scikit-learn', file=sys.stderr)
+    else:
+        params.device = None
```
michael-smirnov, line 205: I think we should check if the device parameter is passed by the user and print a warning that it is useless in that case - for clarity

configs/skl_with_context_config.json
michael-smirnov: Exactly the same what happens if you try to run on a CPU wo DPC++ support - an exception. What is your suggession here?

michael-smirnov: We shall point somewhere that using this config file requires DPC++ support and GPU device on board

runner.py
```diff
@@ -70,6 +70,9 @@ def generate_cases(params):
     parser.add_argument('--report', default=False, action='store_true',
                         help='Create an Excel report based on benchmarks results. '
                              'Need "openpyxl" library')
+    parser.add_argument('--device', default=None, type=str,
```
michael-smirnov, line 73: Why the parameter is duplicated in `bench.py` and `runner.py`?

sklearn_bench/dbscan.py
michael-smirnov: @PetrovKP what about other files that @vlad-nazarov did not touch?

bench.py
PetrovKP: ohh, then I think the host is not needed at all

configs/skl_with_context_config.json
PetrovKP: Ok, probably it's ok

sklearn_bench/dbscan.py
PetrovKP: I will correct if there are such places yet

bench.py
```diff
@@ -197,6 +201,12 @@ def parse_args(parser, size=None, loop_types=(),
         except ImportError:
             print('Failed to import daal4py.sklearn.patch_sklearn.'
                   'Use stock version scikit-learn', file=sys.stderr)
+            params.device = 'host'
+    else:
+        if params.device != 'host':
+            print('Device context not supported without intel optimized version',
```
michael-smirnov, line 207: Device context is not supported for stock scikit-learn. Please use `--no-intel-optimized=False` with f'--device={params.device}' parameter. Fallback to `--device=None`.

configs/skl_with_context_config.json
```diff
@@ -0,0 +1,77 @@
+{
+    "common": {
+        "lib": ["sklearn"],
```
michael-smirnov, line 3: How stock or intel version of sk is specified for this config?

michael-smirnov (changes requested):

PetrovKP (approved):
merge after green CI

configs/skl_with_context_config.json
vlad-nazarov: Do we need to launch the stock sk in this config? Maybe just add flag `no-intel-optimized` before config run? I think it's better to separate patched and unpatched sk launches

michael-smirnov: Ok - probably its better to skip `device != None` branches for stock sklearn

michael-smirnov (approved):

configs/skl_with_context_config.json
vlad-nazarov: Need to add support to skip these cases?

PR #58, opened 2021-03-22
Applied mypy + flake8 + ISort for all files

RukhovichIV:
A first part of the refactoring


PetrovKP (approved):

datasets/load_datasets.py
```diff
@@ -41,7 +40,7 @@ def try_load_dataset(dataset_name, output_directory):
     if dataset_name in dataset_loaders.keys():
         try:
             return dataset_loaders[dataset_name](output_directory)
-        except:
+        except BaseException:
             logging.warning("Internal error loading dataset")
```
Alexsandruss, line 44: In this handling `except BaseException` is not better than `except`. Add traceback in warning or print exception name at least:
```
except BaseException as e:
    logging.warning(f"Internal error loading dataset: {e}")
```

sklearn_bench/df_clsf.py
```diff
@@ -46,7 +43,8 @@
 
 params = bench.parse_args(parser)
 
-from sklearn.ensemble import RandomForestClassifier
+if not params.no_intel_optimized:
+    from sklearn.ensemble import RandomForestClassifier
```
Alexsandruss, line 47: Where is RandomForestClassifier if params.no_intel_optimized=True? Is benchmark predicted to be failed in this case?

sklearn_bench/knn_clsf.py
```diff
@@ -39,7 +37,8 @@
                     help='Distance metric to use')
 params = bench.parse_args(parser)
 
-from sklearn.neighbors import KNeighborsClassifier
+if not params.no_intel_optimized:
+    from sklearn.neighbors import KNeighborsClassifier
```
Alexsandruss, line 41: It is not clear why estimators are imported only in one case

Alexsandruss (changes requested):

sklearn_bench/df_clsf.py
RukhovichIV: My bad. I just wanted to separate this import from others, so we can ISort them

datasets/load_datasets.py
RukhovichIV: done

sklearn_bench/df_clsf.py
RukhovichIV: changed the behaviour. It should work good now

sklearn_bench/knn_clsf.py
RukhovichIV: Also done

Alexsandruss (approved):

PR #62, opened 2021-04-15
Add support proba svc; add pypi info; fix errors

PetrovKP:


KalyanovD (approved):
No concerns on SVM changes. It is better to close Codefactor criticals before merge.

PR #81, opened 2021-07-20
Extend output result & minor fixes

OnlyDeniko:


datasets/loader_regression.py
```diff
@@ -196,9 +197,10 @@ def year_prediction_msd(dataset_dir: Path) -> bool:
     YearPredictionMSD dataset from UCI repository
     https://archive.ics.uci.edu/ml/datasets/yearpredictionmsd
 
-    TaskType:regression
-    NumberOfFeatures:90
-    NumberOfInstances:515345
+    year_prediction_msd x train dataset (463715, 11)
```
RukhovichIV, line 200: Are you sure about 11 features? Do we have an error in this number now?
Can you please mark these datasets as `regression task` as before (line 71)

datasets/loader_multiclass.py
```diff
@@ -204,9 +208,11 @@ def msrank(dataset_dir: Path) -> bool:
     """
     Dataset from szilard benchmarks: https://github.com/szilard/GBM-perf
 
-    TaskType:multiclass
-    NumberOfFeatures:137
-    NumberOfInstances:1.2M
+    Multiclass classification task
```
RukhovichIV, line 211: Could you please calculate the number of classes as well? This might be done i. e. with `np.unique()`

RukhovichIV:
Pep8 is also important. There's an extra space somewhere :)

datasets/loader_regression.py
OnlyDeniko: Mistake fixed

RukhovichIV (approved):

sklearn_bench/dbscan.py
```diff
@@ -40,7 +40,10 @@ def main():
     labels = dbscan.labels_
 
     params.n_clusters = len(set(labels)) - (1 if -1 in labels else 0)
-    acc = davies_bouldin_score(X, labels)
+    try:
+        acc = davies_bouldin_score(X, labels)
+    except Exception:
```
RukhovichIV, line 45: It would be better to have all scoring functions in the same place (bench.py)
And again, such handling is a bad practice

RukhovichIV (changes requested):

sklearn_bench/dbscan.py
```diff
@@ -40,17 +39,17 @@ def main():
...
                        params=params, functions=['DBSCAN'], times=[time],
-                       accuracies=[acc], accuracy_type='davies_bouldin_score',
+                       metrics=[acc], metric_type='davies_bouldin_score',
                        data=[X], alg_instance=dbscan)
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='scikit-learn DBSCAN benchmark')
-    parser.add_argument('-e', '--eps', '--epsilon', type=float, default=10.,
+    parser.add_argument('-e', '--eps', '--epsilon', type=float, default=0.5,
```
PetrovKP, line 52: why did you change it? this will not affect the measurements of the current configs?

sklearn_bench/nusvc.py
```diff
@@ -43,33 +43,44 @@ def main():
 
     if params.probability:
         state_predict = 'predict_proba'
-        accuracy_type = 'log_loss'
         clf_predict = clf.predict_proba
-
-        def metric_call(x, y):
-            return bench.log_loss(x, y)
+        y_proba_train = clf_predict(X_train)
```
PetrovKP, line 47: I think it makes no sense for svm's to display metrics log_loss, roc_auc, because by default he don't return proba

sklearn_bench/nusvc.py
```diff
@@ -43,33 +43,44 @@ def main():
...
-                       alg_instance=clf)
+    test_acc = bench.accuracy_score(y_test, y_pred)
+
+    bench.print_output(
+        library='sklearn',
+        algorithm='nusvc',
+        stages=['training', state_predict],
+        params=params, functions=['NuSVC.fit', f'NuSVC.{state_predict}'],
+        times=[fit_time, predict_train_time],
+        metric_type=['accuracy', 'log_loss', 'roc_auc'],
```
PetrovKP, line 75: Add please n_sv:
clf.n_support_.sum()

sklearn_bench/dbscan.py
OnlyDeniko: It is default value in sklearn

sklearn_bench/nusvc.py
OnlyDeniko: If user want to turn on probability then he will see these metrics

bench.py
```diff
@@ -338,20 +338,47 @@ def columnwise_score(y, yp, score_func):
...
+    if y_pred.shape[1] == 2:  # binary case
+        y_pred = y_pred[:, 1]
+    return sklearn_roc_auc(y_true, y_pred, multi_class=multi_class)
 
-def rmse_score(y, yp):
+
+def rmse_score(y_true, y_pred):
     return columnwise_score(
-        y, yp, lambda y1, y2: float(np.sqrt(np.mean((y1 - y2)**2))))
+        y_true, y_pred, lambda y1, y2: float(np.sqrt(np.mean((y1 - y2)**2))))
```
PetrovKP, line 363: use from sklearn?

bench.py
```diff
@@ -338,20 +338,47 @@ def columnwise_score(y, yp, score_func):
...
+
+
+def davies_bouldin_score(y_true, y_pred):
+    from sklearn.metrics.cluster import davies_bouldin_score as sklearn_dbs
+    y_true = convert_to_numpy(y_true)
+    y_pred = convert_to_numpy(y_pred)
+    try:
+        res = sklearn_dbs(y_true, y_pred)
+    except ValueError:
+        res = "Number of labels is 1"
```
PetrovKP, line 380: for example: y _pred= [1, 1, 1, 1, 1, 1, 1,] ? Or?

sklearn_bench/dbscan.py
PetrovKP: this will not affect the measurements of the current configs?

PetrovKP (approved):
no concerns from my side

bench.py
```diff
@@ -338,20 +338,47 @@ def columnwise_score(y, yp, score_func):
...
+
+def davies_bouldin_score(X, labels):
+    from sklearn.metrics.cluster import davies_bouldin_score as sklearn_dbs
+    X = convert_to_numpy(X)
+    labels = convert_to_numpy(labels)
+    try:
+        res = sklearn_dbs(X, labels)
+    except ValueError as ex:
+        res = ex
+    return res
```
RukhovichIV, line 381: I'm slightly concerned about such exception handling. My opinion is - our own configs shouldn't cause any exceptions.
And I would have rather removed every handling from the code

RukhovichIV (approved):
Great job!
Let's fix the CodeFactor CI asap

Issue #95, opened 2021-10-01
Code quality improvements - adress findings from www.codefactor.io


napetrov:
Please create PRs for addressing code quality issues reported by codefactor scans

https://www.codefactor.io/repository/github/IntelPython/scikit-learn_bench

Some problems are simple fixes and i would expect that they can be fixed across entire repo and not one by one
Other might be super complex and i don't think can be fixed without serious refactoring - code complexity

LyndonFan:
Hi, is this issue available? I can try to work on it!

napetrov:
@LyndonFan  - yes, sure. Please open PRs with fixes

PR #97, opened 2021-10-02
Refactor code

LyndonFan:
Addresses most issues in #95

- Fixed 2 instances of unnecessary else
- Fixed 2 instances of using range(len(...))
- Fixed 1 instance of code complexity by removing unnecessary "is not None" check
- Fixed 1 instance of code security by using requests.get instead of urllib.request.urlretrieve

There is still 1 instance of code complexity issue remaining. I have performed some minor refactoring, but it's still too complex according to codefactor.io.

### DIFF
--- a/daal4py_bench/requirements.txt
+++ b/daal4py_bench/requirements.txt
@@ -3,3 +3,4 @@ pandas < 1.3.0
 daal4py
 openpyxl
 tqdm
+requests
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -75,6 +75,6 @@ jobs:
           addToPath: true
       - script: |
           python -m pip install --upgrade pip setuptools
-          pip install mypy data-science-types requests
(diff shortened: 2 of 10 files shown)


LyndonFan:
The new code has passed all the checks. Would it be possible to get a review?

Alexsandruss (approved):
I see no problems, let's wait for at least one additional approve.

daal4py_bench/requirements.txt
```diff
@@ -3,3 +3,4 @@ pandas < 1.3.0
 daal4py
 openpyxl
 tqdm
+requests
```
napetrov, line 6: Might be missing some context here - why we need requests module? Per my understanding we do not suppose to use it?

LyndonFan: This is to replace the urllib.request.urlretrieve used in load_data. The urllib library failed a check in code factor so I used requests instead.

napetrov (approved):

PR #98, opened 2021-10-12
Update to readme: GPU support

outoftardis:
image


outoftardis:
@vlad-nazarov you ok with merging this?

PR #101, opened 2021-12-21
Add available devices and dtypes selection to runner.py

dmitrii-kriukov:
Add parameters to runner.py that define devices and data types available to run
Unselected devices and data types will be ignored in configs

for example if config contains
`device: ["host", "gpu",  "cpu"]`
and you run
`python runner.py --devices cpu gpu`
cpu and gpu cases will be launched,
but if config contains:
`device: ["host", "gpu"]`
and you run 
`python runner.py --devices cpu gpu`
only  gpu case will be launched 

default values ​​do not limit the available devices and data types

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -44,9 +44,17 @@ if __name__ == '__main__':
                         default='configs/config_example.json',
                         help='The path to a configuration file or '
                              'a directory that contains configuration files')
+    parser.add_argument('--device', '--devices', default='host cpu gpu None', type=str, nargs='+',
+                        choices=('host', 'cpu', 'gpu', 'None'),
+                        help='Availible execution context devices. '
+                        'This parameter only marks devices as available, '
+                        'make sure to add the device to the config file to run it on a specific device')
     parser.add_argument('--dummy-run', default=False, action='store_true',
                         help='Run configuration parser and datasets generation '
                              'without benchmarks running')
+    parser.add_argument('--dtype', '--dtypes', type=str, default="float32 float64", nargs='+',
+                        choices=("float32", "float64"),
+                        help='Available Data types')
     parser.add_argument('--no-intel-optimized', default=False, action='store_true',
                         help='Use Scikit-learn without Intel optimizations')
     parser.add_argument('--output-file', default='results.json',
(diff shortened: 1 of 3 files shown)


runner.py
```diff
@@ -44,9 +44,18 @@ def get_configs(path: Path) -> List[str]:
                         default='configs/config_example.json',
                         help='The path to a configuration file or '
                              'a directory that contains configuration files')
+    parser.add_argument('--device', '--devices', default='host cpu gpu None', type=str, nargs='+',
+                        choices=('host', 'cpu', 'gpu', 'None'),
```
Alexsandruss, line 48: Minor: string
```python
"None"
```
looks confusing because of 
```python 
None
```
existance. May be, switch to `none`?

runner.py
```diff
@@ -44,9 +44,18 @@ def get_configs(path: Path) -> List[str]:
...
+                        help='Availible execution context devices. '
+                        'This parameter only marks devices as available, '
+                        'make sure to add the device to the config file '
+                        'to run it on a specific device')
     parser.add_argument('--dummy-run', default=False, action='store_true',
                         help='Run configuration parser and datasets generation '
                              'without benchmarks running')
+    parser.add_argument('--dtype', '--dtypes', type=str, default="float32 float64", nargs='+',
+                        choices=("float32", "float64"),
```
Alexsandruss, line 57: Out of PR scope, but int32 might be added

runner.py
```diff
@@ -44,9 +44,18 @@ def get_configs(path: Path) -> List[str]:
...
+                        help='Available Data types')
```
Alexsandruss, line 58: Extend description like in `--devices`

runner.py
```diff
@@ -93,6 +102,28 @@ def get_configs(path: Path) -> List[str]:
         for params_set in config['cases']:
             params = common_params.copy()
             params.update(params_set.copy())
+
+            device = []
+            if 'device' not in params:
+                if 'sklearn' in params['lib']:
+                    logging.info('The device parameter value is not defined in config, '
+                                 'none is used')
```
Alexsandruss, line 110: Move up from `if` because `--device` might be added later for xgboost, for example.

Alexsandruss (changes requested):

runner.py
dmitrii-kriukov: For now other configs not support device parameter and such message may be confusing

Alexsandruss (approved):

Pahandrovich (approved):

KalyanovD (approved):

runner.py
samir-nasibli: It is better to add this kind of changes in other PR

samir-nasibli (approved):

PR #103, opened 2021-12-29
Add sizes selection to runner py

dmitrii-kriukov:
Add parameters to runner.py that define sizes of running datasets.
Size parameter may be defined in cases in config file, if size is not defined, case will be launched anyway

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -58,7 +58,7 @@ if __name__ == '__main__':
                         help='Available floating point data types'
                         'This parameter only marks dtype as available, '
                         'make sure to add the dtype parameter to the config file ')
-    parser.add_argument('--size', type=str, default="small medium large", nargs='+',
+    parser.add_argument('--workload-size', type=str, default="small medium large", nargs='+',
                         choices=("small", "medium", "large"),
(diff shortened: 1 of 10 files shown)


vlad-nazarov (approved):

dmitrii-kriukov:
http://intel-ci.intel.com/ec798aa8-05e9-f131-b15b-8cdcd4b723a1

runner.py
```diff
@@ -105,6 +110,11 @@ def get_configs(path: Path) -> List[str]:
             params = common_params.copy()
             params.update(params_set.copy())
 
+            if 'workload-size' in params:
+                if params['workload-size'] not in args.workload_size:
+                    continue
+                del params['workload-size']
```
samir-nasibli, line 116: It is better here to use dictionary own methods to remove instead of using `del` for removing object. It seems that your addition are consistent with other same cases in this script code. This is out of the scope for the current PR, but it would be better to fix this kind of thing in other PR(s).

Alexsandruss (approved):

samir-nasibli (approved):

PR #105, opened 2022-01-20
Configs for xpu: dbscan, kmeans, pca

dmitrii-kriukov:
Configs for dbscan, kmeans, pca
This configs have some problems with float32 in sklearnex, so listed measurements uses float64
image
image
PCA have some issues on gpu

stock vs extension
image
image
image


datasets/loader_classification.py
```diff
@@ -688,6 +722,51 @@ def skin_segmentation(dataset_dir: Path) -> bool:
     return True
 
 
+def cifar(dataset_dir: Path) -> bool:
```
samir-nasibli, line 725: it seems that by the result of the function you will never get a `False` value...

dmitrii-kriukov: The loaders in this file follow the general look of the loaders in the sklearn bench and should return True as a success status. Otherwise it will result in errors here
https://github.com/IntelPython/scikit-learn_bench/blob/master/datasets/load_datasets.py#L78

Pahandrovich (approved):

PR #106, opened 2022-01-21
Configs for xpu: knn and svm

dmitrii-kriukov:
Configs for knn and svm:
image
image
image
image


dmitrii-kriukov:
stock vs sklearnex

image
image
image
image

PR #107, opened 2022-01-27
Report config selector

dmitrii-kriukov:
Add option for report config selection
If report flag is not set, report generator will not be launched
If no config specified default config will be used

`python runner.py --report`
(launch report with default config)
`python runner.py --report report_generator/myconf.json`
(launch report with  custom config)

Added more detailed config for new benchmark cases

### DIFF
--- a/report_generator/xpu_report_gen_config.json
+++ b/report_generator/xpu_report_gen_config.json
@@ -1,0 +1,31 @@
+{
+    "header": [
+        "algorithm",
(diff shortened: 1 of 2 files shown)


dmitrii-kriukov:
http://intel-ci.intel.com/ec82c791-9275-f126-b310-8cdcd4b723a1

vlad-nazarov (approved):

PR #119, opened 2022-09-27
XPU cases for sklearnex DF

samir-nasibli:
The cases will be moved to [skl_xpu_config.json](https://github.com/IntelPython/scikit-learn_bench/blob/master/configs/skl_xpu_config.json)


Alexsandruss:
Config format was changed in https://github.com/IntelPython/scikit-learn_bench/pull/133. These cases will be also covered here.

PR #125, opened 2023-01-31
fixing modelbuilder benchmark for covtype dataset

razdoburdin:
Fix allows to run LGBM model builder benchmark for covtype dataset

### DIFF
--- a/modelbuilders_bench/lgbm_mb.py
+++ b/modelbuilders_bench/lgbm_mb.py
@@ -20,6 +20,7 @@ import os
 import bench
 import daal4py
 import lightgbm as lgbm
+import numpy as np
 
 import modelbuilders_bench.mb_utils as utils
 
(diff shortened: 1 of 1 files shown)


modelbuilders_bench/lgbm_mb.py
```diff
@@ -99,7 +98,7 @@
     if 'cudf' in str(type(y_train)):
         params.n_classes = y_train[y_train.columns[0]].nunique()
     else:
-        params.n_classes = len(np.unique(y_train))
+        params.n_classes = int(max(y_train)) + 1
```
Alexsandruss, line 101: I guess it's safer to add extra inner `if` for case where max of uniques is not equal to length of uniques - 1

Alexsandruss (changes requested):

Alexsandruss:
PEP8 compliance is required

Alexsandruss (approved):

PR #130, opened 2023-03-01
np.float is depricated in numpy 1.24

razdoburdin:
np.float is [depricated](https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations) in numpy 1.24

### DIFF
--- a/modelbuilders_bench/mb_utils.py
+++ b/modelbuilders_bench/mb_utils.py
@@ -23,7 +23,7 @@ def get_accuracy(true_labels, prediction):
     errors = 0
     for i, true_label in enumerate(true_labels):
         pred_label = 0
-        if isinstance(prediction[i], (float, np.single)):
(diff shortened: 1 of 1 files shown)


modelbuilders_bench/mb_utils.py
```diff
@@ -23,7 +23,7 @@ def get_accuracy(true_labels, prediction):
     errors = 0
     for i, true_label in enumerate(true_labels):
         pred_label = 0
-        if isinstance(prediction[i], (float, np.single, np.float)):
+        if isinstance(prediction[i], (float, np.single)):
```
Alexsandruss, line 26: It's better to specify more familiar type names
```suggestion
        if isinstance(prediction[i], (np.float32, np.float64)):
```

Alexsandruss (changes requested):

Alexsandruss (approved):

Issue #75, opened 2021-06-29
Benchmarks silently execute stock version if scikit-learn-intelex is not installed


michael-smirnov:
If I try to run benchmarks with the command like 
```sh
python runner.py --configs configs/skl_xpu_config.json
```
it should run a patched version of scikit-learn algorithms. However, if scikit-learn-intelex package is not installed, benchmarks cannot be patched and should print an error or warning. 

I cannot see the message in https://github.com/IntelPython/scikit-learn_bench/blob/master/bench.py#L204 for some reason.

lulin789:
SKL_Bench_Team1 working on this issue

PR #134, opened 2023-04-28
Fix issue #75: Benchmarks silently execute stock version if scikit-learn-intelex is not installed

lulin789:
Fix log not displayed in bench.py
Before this fix the log isn't displayed in bench.py, because bench.py is
running via subprocess, so the log config set in runner.py has no effect in
bench.py, and the log config in bench.py follows the default strategy
(log_level=WARNING, stream=stderr), so when scikit-learn-intelex is not
installed, the info log can't be displayed, furthermore, all other info logs
in bench.py can't be displayed either.

This fix does the following change:
1. Set a log config in bench.py (log_level=INFO, stream=stderr).
2. Add return_code in the return value of utils.read_output_from_command() to identify if there is any error in subprocess.
3. In runner.py, if no error occured, the log from subprocess in stderr will be printed, otherwise is_success will be set to False and the error message will be displayed through a warning log.

Fixes https://github.com/IntelPython/scikit-learn_bench/issues/75

Co-authored-by: Wu, Zihan <zihan.wu@intel.com>
Signed-off-by: Deng, Lulin <lulin.deng@intel.com>
Signed-off-by: Zhou, Shuangpeng <shuangpeng.zhou@intel.com>
Signed-off-by: Xu, Yanyue <yanyue.xu@intel.com>

### DIFF
--- a/bench.py
+++ b/bench.py
@@ -187,6 +187,8 @@ def parse_args(parser, size=None, loop_types=(),
                         choices=('host', 'cpu', 'gpu', 'none'),
                         help='Execution context device')
 
+    logging.basicConfig(stream=sys.stderr, format='%(levelname)s: %(message)s', level=logging.INFO)
+
     for data in ['X', 'y']:
         for stage in ['train', 'test']:
             parser.add_argument(f'--file-{data}-{stage}',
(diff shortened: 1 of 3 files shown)


samir-nasibli:
@Alexsandruss are we going to integrate this into main branch or we have a fix on your branch?

Alexsandruss:
Benchmarks output warning if sklearnex patching has failed starting from refactoring (https://github.com/IntelPython/scikit-learn_bench/pull/133).

PR #139, opened 2023-09-25
Extratrees inclusion to sklearn_bench

icfaust:
This creates the necessary benchmarks necessary for the ExtraTreesClassifier and ExtraTreesRegressor. It also corrects some errors in the master associated with spelling mistakes or incorrect parameters.

### DIFF
--- a/configs/testing/sklearn.json
+++ b/configs/testing/sklearn.json
@@ -58,6 +58,42 @@
             ],
             "num-trees": 10
         },
+                {
+            "algorithm": "et_clsf",
+            "dataset": [
+                {
+                    "source": "synthetic",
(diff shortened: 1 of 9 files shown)


configs/testing/sklearn.json
```diff
@@ -57,6 +57,42 @@
                 }
             ],
             "num-trees": 10
+        },
+                {
+            "algorithm": "et_clsf",
```
samir-nasibli, line 62: IS it passible to use the same config for RF and ET?

Alexsandruss:
Implemented in refactoring (https://github.com/IntelPython/scikit-learn_bench/pull/133)

PR #143, opened 2023-10-27
Fix LGBM incompatibilities with report_generator

ahuber21:


samir-nasibli (commented):
@ahuber21 what we are going to do with this PR?
What is considered?

ahuber21:
I would still like to merge it. LGBM Model Builder benchmarks are currently broken, and this fix makes them runnable again.

ahuber21:
@Alexsandruss can you take a look here? Are the benchmark reports for model builder results fixed after your big update?

Alexsandruss:
> @Alexsandruss can you take a look here? Are the benchmark reports for model builder results fixed after your big update?

Yes, sklbench report works with LightGBM by default now since it supports sklearn's estimator API.

PR #145, opened 2023-12-06
More specific write access message

ethanglaser:
Associated with task: Fix sklearnex_bench errors if DATASETSROOT is not writeable

- Functionally, the code should already be working as intended - a warning is raised because the path is not writeable, therefore it writes to local path instead. The existing exception is broader so it makes it seem like a problem. Solution is to add a more specific exception for the case where the error is permission denied/not writeable and raise a more specific message. In both cases it's a warning and functionally the same.

### DIFF
--- a/datasets/make_datasets.py
+++ b/datasets/make_datasets.py
@@ -34,7 +34,7 @@ def try_gen_dataset(args, folder):
         else:
             raise ValueError(f'{args.type} is unknown dataset type')
         return True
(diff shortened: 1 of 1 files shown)


ethanglaser:
http://intel-ci.intel.com/ee943eff-2a76-f100-9ac0-a4bf010d0e2e

Alexsandruss:
Not relevant after refactoring. New caching scheme requires write access.

PR #153, opened 2024-08-08
update n_init for kmeans

md-shafiul-alam:
- For kmeans++, the default n_init=1 https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/cluster/_kmeans.py#L346. Probably makes sense to benchmark the default behavior.

### DIFF
--- a/configs/regular/kmeans.json
+++ b/configs/regular/kmeans.json
@@ -21,6 +21,11 @@
         "sklearn kmeans parameters": {
             "algorithm": { "estimator_params": { "init": "k-means++", "algorithm": "lloyd" } }
         },
+        "cuml kmeans parameters": {
+            "algorithm": {
+                "estimator_params": { "init": "scalable-k-means++" }
+            }
+        },
         "kmeans datasets": [
             {
                 "data": [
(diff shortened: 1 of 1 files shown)


Alexsandruss:
@md-shafiul-alam 

> Change normalize to false

But why? KMeans inertia is sensitive to normalization

md-shafiul-alam:
> @md-shafiul-alam
> 
> > Change normalize to false
> 
> But why? KMeans inertia is sensitive to normalization

Only reason is that stock scikit learn does not have any option for normalization.

Alexsandruss:
`data:preprocessing_kwargs:normalize` is not argument of an estimator

md-shafiul-alam:
> `data:preprocessing_kwargs:normalize` is not argument of an estimator

I understand, but is there a specific reason we normalize by default? To me seems like a redundant step.

Alexsandruss (changes requested):
> > `data:preprocessing_kwargs:normalize` is not argument of an estimator
> 
> I understand, but is there a specific reason we normalize by default? To me seems like a redundant step.

Normalization is required for algorithms sensitive to it for better quality metrics and representation of user cases.

md-shafiul-alam:
@Alexsandruss Do you suggest any more changes?

Alexsandruss (approved):

PR #156, opened 2024-08-23
Move summary sheet to first place in report

Alexsandruss:
Fix for summary page without index: move it from last place to first in workbook.


icfaust (approved):
Looks like openpyxl documentation doesn't list if its 0 indexed or 1 indexed, looks right to me though. Approved.

PR #160, opened 2024-09-20
Add incremental algorithms support

olegkkruglov:
# Description

* Added support of incremental algorithms
* Added support of multidataset run for incremental algorithms
* Added config example for introduced functionality
* Refactor dataset loading to make possible measuring scenarios with multiple dataset loads.


sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -345,23 +347,27 @@ def ndarray_function(x, y):
                     x[i * batch_size : (i + 1) * batch_size],
                     y[i * batch_size : (i + 1) * batch_size],
                 )
+            estimator_instance._onedal_finalize_fit()
```
Alexsandruss, line 350: Estimator instance is generic sklearn-like estimator without necessary finalize_fit
```suggestion
            if hasattr(estimator_instance, "_onedal_finalize_fit"):
                estimator_instance._onedal_finalize_fit()
```

Alexsandruss (changes requested):

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: but if we are here it means that estimator has partial_fit method and it implies that it has finalize_fit method

sklbench/report/implementation.py
```diff
@@ -239,6 +239,7 @@ def get_result_tables_as_df(
     bench_cases = pd.DataFrame(
         [flatten_dict(bench_case) for bench_case in results["bench_cases"]]
     )
+    bench_cases = bench_cases.map(lambda x: str(x) if not isinstance(x, Hashable) else x)
```
Alexsandruss, line 242: What is non-hashable object you are trying to convert?

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -414,12 +424,28 @@ def measure_sklearn_estimator(
...
+                    if batch_size is None:
+                        if num_batches is None:
+                            num_batches = 5
+                        batch_size = (
+                            data_args[0].shape[0] + num_batches - 1
+                        ) // num_batches
+                    if num_batches is None:
+                        num_batches = (
+                            data_args[0].shape[0] + batch_size - 1
+                        ) // batch_size
```
Alexsandruss, line 441: Batching is regulated by `batch_size` only.
All default parameters should be defined in `get_bench_case_value`.

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -334,7 +334,9 @@ def verify_patching(stream: io.StringIO, function_name) -> bool:
     return acceleration_lines > 0 and fallback_lines == 0
 
 
-def create_online_function(method_instance, data_args, batch_size):
+def create_online_function(
+    estimator_instance, method_instance, data_args, num_batches, batch_size
```
Alexsandruss, line 338: Where `num_batches` is used in this function?

configs/incremental.json
```diff
@@ -0,0 +1,96 @@
...
+            "data": {
+                "source": "make_blobs",
+                "generation_kwargs": {
+                    "centers": 1,
+                    "n_samples": 1000,
+                    "n_features": [16, 64]
+                },
+                "split_kwargs": {"ignore": true}
+            }
+        },
```
Alexsandruss, line 50: What is the point of different datasets in example config?

Alexsandruss (changes requested):

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: > Batching is regulated by `batch_size` only.
why do you think this is strict rule?

sklbench/report/implementation.py
olegkkruglov: basic statistics result_options parameter is a list

configs/incremental.json
olegkkruglov: why not? it is just an example

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: fixed

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: > All default parameters should be defined in `get_bench_case_value`.

here num_batches is set to default only if both num_batches and batch_size are None. such logic can't be implemented via get_bench_case_value as far as I understand

configs/incremental.json
Alexsandruss: This example doesn't show any extra feature except incremental algs. Datasets are better shown in sklearn_example.json

Alexsandruss: It is better to reduce datasets like it's done in spmd example

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -334,34 +334,43 @@ def verify_patching(stream: io.StringIO, function_name) -> bool:
...
-    n_batches = data_args[0].shape[0] // batch_size
+def create_online_function(
+    estimator_instance, method_instance, data_args, num_batches, batch_size
+):
 
     if "y" in list(inspect.signature(method_instance).parameters):
 
         def ndarray_function(x, y):
-            for i in range(n_batches):
+            for i in range(num_batches):
```
Alexsandruss, line 344: Leave old simple logic with `batch_size` only.

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -414,12 +423,28 @@ def measure_sklearn_estimator(
...
                     else:
                         data_args = (x_test,)
-                batch_size = get_bench_case_value(
-                    bench_case, f"algorithm:batch_size:{stage}"
-                )
-                if batch_size is not None:
+
+                if method == "partial_fit":
+                    num_batches = get_bench_case_value(bench_case, "data:num_batches")
+                    batch_size = get_bench_case_value(bench_case, "data:batch_size")
```
Alexsandruss, line 429: Instead of separate branch for `partial_fit`, extend mechanism of `online_inference_mode` to partial fitting too.

Alexsandruss (changes requested):

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: why?

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: could you provide the exact link to implementation of this mechanism? i was not able to find the usage of this parameter, just see its setting in the config.

configs/incremental.json
```diff
@@ -0,0 +1,96 @@
...
+        "covariance": {
+            "algorithm": [
+                {
+                    "estimator": "IncrementalEmpiricalCovariance",
+                    "library": "sklearnex.covariance",
```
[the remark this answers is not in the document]
olegkkruglov: here it doesn't work since currently covariance module is not in `sklearnex.__init__.py`. This is fixed here: https://github.com/intel/scikit-learn-intelex/pull/2072 and the config can be updated after the fix will reach pip version of sklearnex.

sklbench/benchmarks/sklearn_estimator.py
Alexsandruss: Actrually, `online_inference_mode` was removed as unnecessary before merge of refactor branch. This mode is enabled by `batch_size != None` only.

sklbench/benchmarks/sklearn_estimator.py
Alexsandruss: Why change? It overcomplicates data slicing with extra parameter checks and calculations, also, it is more common to know batch size before `partial_fit` call in real world cases.

configs/incremental.json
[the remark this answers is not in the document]
Alexsandruss: Create ticket for future update.

configs/incremental.json
```diff
@@ -0,0 +1,74 @@
...
+                    "num_batches": {"training": 2}
+                }
+            ]
+        },
+        "basic_statistics": {
+            "algorithm": [
+                {
+                    "estimator": "IncrementalBasicStatistics",
+                    "library": "sklearnex",
+                    "num_batches": {"training": 2}
```
Alexsandruss, line 44:
```suggestion
                    "library": "sklearnex"
```

configs/incremental.json
```diff
@@ -0,0 +1,74 @@
...
+                }
+            ]
+        },
+        "linear_regression": {
+            "algorithm": [
+                {
+                    "estimator": "IncrementalLinearRegression",
+                    "library": "sklearnex",
+                    "num_batches": {"training": 2}
```
Alexsandruss, line 53:
```suggestion
                    "library": "sklearnex"
```

configs/incremental.json
```diff
@@ -0,0 +1,74 @@
...
+                }
+            ]
+        },
+        "pca": {
+            "algorithm": [
+                {
+                    "estimator": "IncrementalPCA",
+                    "library": "sklearnex.preview",
+                    "num_batches": {"training": 2}
```
Alexsandruss, line 62:
```suggestion
                    "library": "sklearnex.preview"
```

configs/incremental.json
```diff
@@ -0,0 +1,74 @@
...
+            ]
+        }
+    },
+    "TEMPLATES": {
+        "covariance": {"SETS": ["common", "covariance", "unlabeled dataset"]},
+        "linear_regression": {
+            "SETS": ["common", "linear_regression", "labeled dataset"]
+        },
+        "pca": {"SETS": ["common", "pca", "unlabeled dataset"]}
+    }
```
Alexsandruss, line 73: Basic statistic parameters are not used.

configs/incremental.json
Alexsandruss, line 1: `sklearnex_incremental_example.json` file name is more clear.

sklbench/benchmarks/sklearn_estimator.py
Alexsandruss: You can split batch size into two for training and inference.

configs/incremental.json
olegkkruglov: here is the similar problem as with covariance. https://github.com/intel/scikit-learn-intelex/pull/2022
it fails if intelex version without this fix is used. i'll add this to the ticket too to fix when pip version of intelex will be updated.

configs/incremental.json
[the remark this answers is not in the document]
olegkkruglov: ticket 8528

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: > Actually, `online_inference_mode` was removed as unnecessary before merge of refactor branch.
what should I extend then?

sklbench/benchmarks/sklearn_estimator.py
olegkkruglov: > Why change?

Adding new feature which can be useful.

> It overcomplicates data slicing with extra parameter checks and calculations

It costs nothing. And doing calculations in the code is better than doing them in calculator before running benchmarks.

> it is more common to know batch size before `partial_fit` call in real world cases.

But while doing benchmarking it is not less common (I'd say even more) when the user wants to specify exact number of partial_fit calls.

ethanglaser: I don't think it makes sense to have both, since one would depend on the other

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -334,34 +334,43 @@ def verify_patching(stream: io.StringIO, function_name) -> bool:
...
                 method_instance(x[i * batch_size : (i + 1) * batch_size])
+            if hasattr(estimator_instance, "_onedal_finalize_fit"):
+                estimator_instance._onedal_finalize_fit()
 
         def dataframe_function(x):
-            for i in range(n_batches):
+            for i in range(num_batches):
                 method_instance(x.iloc[i * batch_size : (i + 1) * batch_size])
+            if hasattr(estimator_instance, "_onedal_finalize_fit"):
+                estimator_instance._onedal_finalize_fit()
```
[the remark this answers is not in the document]
olegkkruglov: finalize_fit is called if any result attribute has been called. only partial_fit would be measured here without this call

sklbench/benchmarks/sklearn_estimator.py
ethanglaser: Upon further investigation and using this branch, I think the setup here makes sense

configs/sklearnex_incremental_example.json
```diff
@@ -0,0 +1,65 @@
...
+                    "noise": 2.0
+                }
+            }
+        },
+        "covariance": {
+            "algorithm": [
+                {
+                    "estimator": "IncrementalEmpiricalCovariance",
+                    "library": "sklearnex.covariance",
+                    "estimator_methods": {"training": "partial_fit"},
```
ethanglaser, line 34: is there a reason estimator_methods is only specified for one algo?

sklbench/benchmarks/sklearn_estimator.py
ethanglaser: It just allows user to specify either num_batches or batch_size in the config. Original usage is unimpacted. I have no objections to this

configs/README.md
```diff
@@ -117,6 +117,7 @@ Configs have the three highest parameter keys:
 |:---------------|:--------------|:--------|:------------|
 | `algorithm`:`estimator` | None |  | Name of measured estimator. |
 | `algorithm`:`estimator_params` | Empty `dict` |  | Parameters for estimator constructor. |
+| `algorithm`:`training`:`num_batches` | 5 |  | Number of batches to benchmark `partial_fit` function, using batches the size of number of samples specified (not samples divided by `num_batches`). For incremental estimators only. |
```
Alexsandruss, line 120: 1. Why same is not applied to inference?
2. Wrong order of keys:
```suggestion
| `algorithm`:`num_batches`:`training` | 5 |  | Number of batches to benchmark `partial_fit` function, using batches the size of number of samples specified (not samples divided by `num_batches`). For incremental estimators only. |
```

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -74,7 +74,7 @@ def get_estimator(library_name: str, estimator_name: str):
 def get_estimator_methods(bench_case: BenchCase) -> Dict[str, List[str]]:
     # default estimator methods
     estimator_methods = {
-        "training": ["fit"],
+        "training": ["partial_fit", "fit"],
```
Alexsandruss, line 77: I think `partial_fit` should be explicitly requested in config since most of incremental estimators can work in both modes:
```suggestion
        "training": ["fit"],
```

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -417,9 +418,17 @@ def measure_sklearn_estimator(
                 batch_size = get_bench_case_value(
                     bench_case, f"algorithm:batch_size:{stage}"
                 )
-                if batch_size is not None:
```
Alexsandruss, line 420: Batch size setting is required by inference measurements.

Alexsandruss (changes requested):

PR #171, opened 2025-03-06
FIX: Fix error in report when sklearnex has more parameters than sklearn

david-cortes-intel:
## Description

Currently, when creating a report comparing sklearn vs. sklearnex, if some algorithm has more parameters in sklearnex  than in sklearn (such as tree-based algorithms like `ExtraTreesClassifier`), an error will occur when trying to generate the excel report from the JSONs, due to being unable to find those sklearnex-specific columns in the data from sklearn.

This PR fixes the issue by dropping all parameter columns that are not present in data from all the compared libraries, but I am not sure that this wouldn't break some other use-case, so please give it a thorough review.

---

PR should start as a draft, then move to ready for review state after CI is passed and all applicable checkboxes are closed.
This approach ensures that reviewers don't spend extra time asking for regular requirements.

You can remove a checkbox as not applicable only if it doesn't relate to this PR in any way.

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/intel/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.

**Testing**

### DIFF
--- a/sklbench/report/implementation.py
+++ b/sklbench/report/implementation.py
@@ -16,8 +16,10 @@
 
 import argparse
 import json
+from functools import reduce
 from typing import Dict, List
 
+import numpy as np
 import openpyxl as xl
 import pandas as pd
 from openpyxl.formatting.rule import ColorScaleRule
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
Looks like this PR doesn't work in all cases. Sometimes it leaves rows where there are times for both sklearn and sklearnex but no comparison metrics (e.g. time improvement) calculated for them.

PR #208, opened 2026-04-13
Revise SVR dataset scope and address uncaught errors

ethanglaser:
## Description

- Removes year_prediction_msd dataset from SVR/NuSVR benchmarks, reducing overall CPU runtime benchmark time from 5-8 hours (http://intel-ci.intel.com/f1309bac-cb88-f1b6-a450-a4bf010d0e2d) to 1-1.5 hours (http://intel-ci.intel.com/f13777e8-38d3-f195-8d41-a4bf010d0e2d)
- Revises measurement.py to resolve uncaught error in sklbench report generation CI `AttributeError: 'float' object has no attribute 'sort'`
- Update downloaders.py to avoid using deprecated array arg
- Update uploadeders.py to only use to_dense if not already dense (otherwise error occurs)

CI status: http://intel-ci.intel.com/f13777e8-38d3-f195-8d41-a4bf010d0e2d

---

<details>
<summary>Checklist:</summary>

- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [X] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [X] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [X] I have run it locally and tested the changes extensively.
- [X] All CI jobs are green or I have provided justification why they aren't.
- [X] I have extended testing suite if new functionality was introduced in this PR.

</details>

### DIFF
--- a/sklbench/datasets/loaders.py
+++ b/sklbench/datasets/loaders.py
@@ -450,7 +450,7 @@ def load_codrnanorm(
     data_name: str, data_cache: str, raw_data_cache: str, dataset_params: Dict
 ) -> Tuple[Dict, Dict]:
     def transform_x_y(x, y):
-        x = pd.DataFrame(x.todense())
+        x = pd.DataFrame(x.todense() if hasattr(x, "todense") else x)
         y = y.astype("int")
         y[y == -1] = 0
         return x, y
--- a/sklbench/datasets/loaders.py
+++ b/sklbench/datasets/loaders.py
@@ -450,7 +450,7 @@ def load_codrnanorm(
     data_name: str, data_cache: str, raw_data_cache: str, dataset_params: Dict
 ) -> Tuple[Dict, Dict]:
     def transform_x_y(x, y):
-        x = pd.DataFrame(x.todense() if hasattr(x, "todense") else x)
+        x = pd.DataFrame(x)
         y = y.astype("int")
         y[y == -1] = 0
         return x, y
--- a/sklbench/utils/measurement.py
+++ b/sklbench/utils/measurement.py
@@ -65,22 +65,31 @@ def enrich_metrics(
     """Transforms raw performance and other results into aggregated metrics"""
     # time metrics
     res = bench_result.copy()
-    mean, std = box_filter(res["time[ms]"])
-    if include_performance_stability_metrics:
+    if isinstance(res["time[ms]"], list):
+        mean, std = box_filter(res["time[ms]"])
+        if include_performance_stability_metrics:
+            res.update(
+                {
+                    "1st run time[ms]": res["time[ms]"][0],
+                    "1st-mean run ratio": res["time[ms]"][0] / mean,
+                }
+            )
         res.update(
             {
-                "1st run time[ms]": res["time[ms]"][0],
-                "1st-mean run ratio": res["time[ms]"][0] / mean,
+                "time[ms]": mean,
+                "time CV": std / mean,  # Coefficient of Variation
             }
         )
-    res.update(
-        {
-            "time[ms]": mean,
-            "time CV": std / mean,  # Coefficient of Variation
-        }
-    )
+    else:
+        # already aggregated (e.g. from a baseline file)
+        mean = res["time[ms]"]
+        std = res.get("time std[ms]", 0.0)
+        if mean != 0:
+            res["time CV"] = std / mean
+        else:
+            res["time CV"] = 0.0
     cost = res.get("cost[microdollar]", None)
-    if cost:
+    if cost and isinstance(cost, list):
         res["cost[microdollar]"] = box_filter(res["cost[microdollar]"])[0]
     batch_size = res.get("batch_size", None)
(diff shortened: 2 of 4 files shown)


david-cortes-intel:
@avolkov-intel I recall you also wanted to make changes like these - would this PR create some conflict later on?

sklbench/utils/measurement.py
```diff
@@ -65,22 +65,31 @@ def enrich_metrics(
...
+                    "1st run time[ms]": res["time[ms]"][0],
+                    "1st-mean run ratio": res["time[ms]"][0] / mean,
+                }
+            )
         res.update(
             {
-                "1st run time[ms]": res["time[ms]"][0],
-                "1st-mean run ratio": res["time[ms]"][0] / mean,
+                "time[ms]": mean,
+                "time CV": std / mean,  # Coefficient of Variation
```
david-cortes-intel, line 80: It's very confusing to have a measurement called "time CV" reflecting the coefficient of variation when we also have procedures doing cross validation there. Also note that this is not actually the coefficient of variation due to the "boxed" methodology.

Perhaps could output just the standard deviation and name it 'std[ms]'. Or otherwise maybe could name it "std to mean (ratio)".

sklbench/utils/measurement.py
```diff
@@ -65,22 +65,31 @@ def enrich_metrics(
...
     res = bench_result.copy()
-    mean, std = box_filter(res["time[ms]"])
-    if include_performance_stability_metrics:
+    if isinstance(res["time[ms]"], list):
+        mean, std = box_filter(res["time[ms]"])
+        if include_performance_stability_metrics:
+            res.update(
+                {
+                    "1st run time[ms]": res["time[ms]"][0],
+                    "1st-mean run ratio": res["time[ms]"][0] / mean,
```
Vika-F, line 74: Is it possible to have mean == 0 here?
```suggestion
                    "1st-mean run ratio": res["time[ms]"][0] / mean if mean != 0 else res["time[ms]"][0],
```

ethanglaser: If the mean runtime is 0 then we have larger problems than a divide by zero. But I did not update the logic here (see [main branch](https://github.com/IntelPython/scikit-learn_bench/blob/main/sklbench/utils/measurement.py#L73)), the diff only shows because I added an additional condition (`if isinstance(res["time[ms]"], list):`). There is no diff within this condition.

sklbench/utils/measurement.py
ethanglaser: I don't disagree with you, in fact for the large-scale branch that we have for multi-gpu we use [std instead of CV](https://github.com/IntelPython/scikit-learn_bench/blob/large-scale/sklbench/utils/measurement.py#L56). But I suggest revising this in a different PR as I did not change the logic here - the only reason the diff shows is because there was a condition added. Scope of this PR is resolving slowdowns/errors/failures and improving stability of CI jobs.

sklbench/datasets/loaders.py
```diff
@@ -450,7 +450,7 @@ def load_codrnanorm(
     data_name: str, data_cache: str, raw_data_cache: str, dataset_params: Dict
 ) -> Tuple[Dict, Dict]:
     def transform_x_y(x, y):
-        x = pd.DataFrame(x.todense())
+        x = pd.DataFrame(x.todense() if hasattr(x, "todense") else x)
```
avolkov-intel: For which cases do we need this? Should we show a warning in case we can't convert to dense?

sklbench/utils/measurement.py
```diff
@@ -65,22 +65,31 @@ def enrich_metrics(
...
+        # already aggregated (e.g. from a baseline file)
+        mean = res["time[ms]"]
+        std = res.get("time std[ms]", 0.0)
+        if mean != 0:
+            res["time CV"] = std / mean
+        else:
+            res["time CV"] = 0.0
     cost = res.get("cost[microdollar]", None)
-    if cost:
+    if cost and isinstance(cost, list):
```
avolkov-intel, line 92: What happens if cost is not a list?

sklbench/datasets/loaders.py
ethanglaser: I believe this was necessary in the case where x was already dense (ie no longer a sparse array) and thus did not have a todense() attribute function. But I can take a closer look.

sklbench/utils/measurement.py
ethanglaser: The scalar value will be used as is

david-cortes-intel (approved):
LGTM, but let's see if there's any further comments from @avolkov-intel .

avolkov-intel (approved):

PR #209, opened 2026-04-23
Improve sklearnex relative perf in sklearn_example.json

ethanglaser:
## Description

Before:
image

After:
image

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/configs/sklearn_example.json
+++ b/configs/sklearn_example.json
@@ -13,7 +13,7 @@
                 "source": "make_blobs",
                 "generation_kwargs": {
                     "centers": 2,
-                    "n_samples": 1000,
+                    "n_samples": 5000,
                     "n_features": [16, 64]
                 },
                 "split_kwargs": { "ignore": true }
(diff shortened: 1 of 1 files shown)


configs/sklearn_example.json
```diff
@@ -13,7 +13,7 @@
                 "source": "make_blobs",
                 "generation_kwargs": {
                     "centers": 2,
-                    "n_samples": 1000,
+                    "n_samples": 5000,
                     "n_features": [16, 64]
```
david-cortes-intel, line 17: Do we need the case with 16 features?

configs/sklearn_example.json
```diff
@@ -65,21 +69,19 @@
                     "estimator": ["KNeighborsClassifier", "KNeighborsRegressor"],
                     "estimator_params": { "n_neighbors": 5, "algorithm": "brute" }
```
avolkov-intel: I suggest trying to use algorithm 'kd_tree' performance might look better

configs/sklearn_example.json
```diff
@@ -33,29 +33,33 @@
...
-                    "n_samples": 1000,
-                    "n_features": 10,
-                    "n_informative": "[SPECIAL_VALUE]0.75"
+                    "n_samples": 10000,
+                    "n_features": 100,
+                    "n_informative": 50,
+                    "noise": 0.1
                 },
-                "split_kwargs": { "test_size": 0.1 }
+                "split_kwargs": { "test_size": 0.2 }
```
avolkov-intel, line 45: Let's increase the portion of test data to get better perf numbers for predict. I think it's possible to get better numbers for LinearRegression predict

configs/sklearn_example.json
```diff
@@ -13,7 +13,7 @@
                 "source": "make_blobs",
                 "generation_kwargs": {
                     "centers": 2,
-                    "n_samples": 1000,
+                    "n_samples": 5000,
                     "n_features": [16, 64]
                 },
                 "split_kwargs": { "ignore": true }
```
avolkov-intel, line 19: Let's specify the portion of test data to improve perf of KMeans predict

ethanglaser:
@avolkov-intel with this diff:
image
I get this result:
image

Seems like kd_tree does have better speedup for predict, but poor for fit (which matters for kd_tree) - for brute there is negligible compute happening in fit so that being red is not really a problem, therefore I am not sure kd_tree is better overall. LinReg predict is marginally better - I think it may be tough to get significant speedups here. And as for kmeans, I think part of the example config is to show off different features of the config, one of which is ignoring like it has currently - so we can add it or not.

avolkov-intel:
> @avolkov-intel with this diff: image I get this result: image
> 
> Seems like kd_tree does have better speedup for predict, but poor for fit (which matters for kd_tree) - for brute there is negligible compute happening in fit so that being red is not really a problem, therefore I am not sure kd_tree is better overall. LinReg predict is marginally better - I think it may be tough to get significant speedups here. And as for kmeans, I think part of the example config is to show off different features of the config, one of which is ignoring like it has currently - so we can add it or not.

I guess let's keep kd_tree for KNN and ignore the rest of the suggestions

configs/sklearn_example.json
ethanglaser: I think this may have been originally added to demonstrate to users that configs can take lists instead of just individual values - so I'd be in favor of keeping it as is

david-cortes-intel (approved):

avolkov-intel (approved):

PR #210, opened 2026-05-06
Update normalization parameters and add estimator params validation

avolkov-intel:
## Description

- Add different normalization options
- Remove implicit normalization from loaders
- Add estimator parameters validation

Estimator parameters validation can be useful in case you want to override some parameter like n_jobs using -p algorithm:estimator_params:n_jobs=64 for the benchmarks. Currently it would fail since some estimators don't support n_jobs and you would need to run the benchmarks separately. In this approach this parameter will be simply ignored by estimators that don't support it and warning will be shown.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/configs/common/knn.json
+++ b/configs/common/knn.json
@@ -8,7 +8,7 @@
                 }
             },
             "data": {
-                "preprocessing_kwargs": { "normalize": true }
+                "preprocessing_kwargs": { "normalize": "standard" }
             }
         },
         "sklearn knn parameters": {
--- a/configs/regular/logreg.json
+++ b/configs/regular/logreg.json
@@ -61,8 +61,7 @@
             {
                 "data": {
                     "dataset": "hepmass",
-                    "split_kwargs": { "train_size": 0.1, "test_size": null },
-                    "preprocessing_kwargs": { "normalize": "standard" }
+                    "split_kwargs": { "train_size": 0.1, "test_size": null }
                 },
                 "algorithm": { "estimator_params": {"C": 1e-5} }
             },
--- a/configs/regular/logreg.json
+++ b/configs/regular/logreg.json
@@ -61,21 +61,24 @@
             {
                 "data": {
                     "dataset": "hepmass",
-                    "split_kwargs": { "train_size": 0.1, "test_size": null }
+                    "split_kwargs": { "train_size": 0.1, "test_size": null },
+                    "preprocessing_kwargs": { "normalize": "standard" }
                 },
                 "algorithm": { "estimator_params": {"C": 1e-5} }
             },
             {
                 "data": {
                     "dataset": "cifar",
-                    "split_kwargs": { "train_size": 0.1, "test_size": null }
+                    "split_kwargs": { "train_size": 0.1, "test_size": null },
+                    "preprocessing_kwargs": { "normalize": "mean" }
                 },
                 "algorithm": { "estimator_params": {"C": 1e-9} }
             },
             {
                 "data": {
                     "dataset": "gisette",
-                    "split_kwargs": { "train_size": 2000, "test_size": null }
+                    "split_kwargs": { "train_size": 2000, "test_size": null },
+                    "preprocessing_kwargs": { "normalize": "standard" }
                 },
                 "algorithm": { "estimator_params": {"C": 1e1} }
             }
--- a/configs/regular/svm.json
+++ b/configs/regular/svm.json
@@ -15,7 +15,7 @@
                 "algorithm": { "estimator_params": { "C": 100.0, "kernel": "rbf" } }
             },
             {
-                "data": { "dataset": "gisette", "preprocessing_kwargs": { "normalize": null } },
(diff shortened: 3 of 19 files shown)


configs/regular/logreg.json
```diff
@@ -61,21 +61,24 @@
             {
                 "data": {
                     "dataset": "hepmass",
-                    "split_kwargs": { "train_size": 0.1, "test_size": null }
+                    "split_kwargs": { "train_size": 0.1, "test_size": null },
```
ethanglaser: Seems like this is the only case where benchmark behavior changes - is it intended?

configs/weekly/dbscan.json
```diff
@@ -4,14 +4,19 @@
         "high-load dbscan datasets": {
             "data": [
                 {
-                    "dataset": ["cifar", "road_network", "covtype"],
+                    "dataset" : "cifar",
+                    "split_kwargs": { "ignore" : true },
```
ethanglaser:
```suggestion
                    "dataset": "cifar",
                    "split_kwargs": { "ignore": true },
```

configs/regular/svm.json
```diff
@@ -30,7 +30,7 @@
                 "data": {
                     "dataset": "mnist",
                     "split_kwargs": { "train_size": 20000, "test_size": null },
-                    "preprocessing_kwargs": { "normalize": false }
+                    "preprocessing_kwargs": {"normalize" : null}
```
ethanglaser:
```suggestion
                    "preprocessing_kwargs": { "normalize": null }
```
Minor adjustment but looks like spaces have been added before colons in a few places throughout the configs

configs/regular/logreg.json
avolkov-intel: I think it was done for a reason but let me check the convergence for both options

configs/common/knn.json
```diff
@@ -8,7 +8,7 @@
                 }
             },
             "data": {
-                "preprocessing_kwargs": { "normalize": true }
+                "preprocessing_kwargs": { "normalize": "standard" }
```
david-cortes-intel, line 11: Would the jsons with the datasets be able to override this? For example:
https://github.com/avolkov-intel/scikit-learn_bench/blob/a2a75152d6bae2cf3bcdf350125369b04f289b33/configs/regular/knn.json#L8

It's not clear from the docs how it would work when specified more than once.

requirements.txt
```diff
@@ -6,6 +6,7 @@ pandas
 tabulate
 fastparquet
 h5py
+openml
```
david-cortes-intel, line 9: Looks like you need to rebase from the current main.

sklbench/datasets/common.py
```diff
@@ -219,9 +224,18 @@ def preprocess_x(
...
+        if normalize == "standard":
+            scaler = StandardScaler(with_mean=True, with_std=True)
+        elif normalize == "mean":
+            scaler = StandardScaler(with_mean=True, with_std=False)
+        elif normalize == "minmax":
+            scaler = MinMaxScaler(feature_range=(0, 1))
+        else:
+            logger.warning(f'Unknown "{normalize}" normalization type.')
+        if scaler is not None:
+            return pd.DataFrame(scaler.fit_transform(x), columns=x.columns, index=x.index)
```
david-cortes-intel: Wouldn't this make it ignore `return_type == np.ndarray`?

requirements.txt
avolkov-intel: It's rebased I just added openml to general requirements.txt file. Currently it's only added to sklearn specific requirements

configs/regular/logreg.json
avolkov-intel: Essentially, there's no effect on the result: accuracy and number of iterations stays the same

sklbench/datasets/common.py
avolkov-intel: Currently it works correctly for all return_types as intermediate data is always represented in pandas format. However, this conversion is indeed redundant if return_type is not a pandas dataframe

configs/common/knn.json
avolkov-intel: As I understand the logic of the benchmarks it's possible to override the parameter value on the lower level which is the case for all parameters not only normalization. I double checked this for this concrete case, normalize parameter is overriden. This is documented at the bottom of this file: https://github.com/IntelPython/scikit-learn_bench/tree/main/docs 

I think that's design pattern of the sklbench, if we want to document it more properly it's better to do this in separate PR

sklbench/datasets/common.py
david-cortes-intel: Is that because it then goes through train_test_split? Isn't that step optional?

ethanglaser (approved):
LGTM once David's comments are resolved. We should add a json linter step to CI and unify formatting in sklbench configs.

david-cortes-intel (approved):
LGTM, but could also change array.values to array.to_numpy().

david-cortes-intel (approved):

PR #212, opened 2026-05-13
Add throughput benchmarking mode for multi-instance workload simulation

avolkov-intel:
## Description

Adds a new `--throughput-mode` to scikit-learn_bench that simulates realistic cloud/server workloads by running multiple synchronized benchmark instances in parallel, each pinned to its own set of CPU cores via `numactl`.

This addresses the overparallelization problem on large servers (e.g., 96+ cores) where single-process benchmarks don't reflect real usage patterns. In practice, such servers run many concurrent workloads each using 8-16 cores.

## How it works

1. For each benchmark case, the orchestrator launches N instances as subprocesses with `numactl --physcpubind=X-Y --localalloc`
2. Each instance independently loads data, creates the estimator, and warms up
3. A TCP socket barrier synchronizes all instances before measurement begins (<0.2ms sync spread)
4. Each instance runs the task (training, then inference) repeatedly for a configurable duration
5. Results are aggregated: per-instance stats + total throughput across all instances

## New CLI arguments

| Argument | Default | Description |
|----------|---------|-------------|
| `--throughput-mode` | off | Enable throughput mode |
| `--num-instances` | — | Number of parallel instances |
| `--cores-per-instance` | — | CPU cores per instance |
| `--measurement-duration` | 60s | Duration per measurement stage |
| `--throughput-full-logs` | off | Store raw per-iteration timing arrays |
| `--emergency-timeout` | 3600s | Safety timeout for subprocesses |

Parameters can also be set per-case in JSON configs under `bench:` namespace.

## Output

Per-instance statistics: mean, std, median, p01/p25/p75/p99 duration, first iteration time, throughput.  
Aggregate: total iterations, total throughput, mean/std/min/max per instance.  
Quality metrics (accuracy, RMSE, etc.) from the final fitted model.

## Files added
- `sklbench/utils/core_assignment.py` — NUMA-aware core partitioning
- `sklbench/benchmarks/throughput_worker.py` — subprocess worker
- `sklbench/runner/throughput.py` — orchestrator with TCP barrier
- `configs/throughput/example.json` — example config

## Files modified
- `sklbench/runner/arguments.py` — new CLI args
- `sklbench/runner/implementation.py` — conditional branch for throughput mode
- `sklbench/utils/measurement.py` — graceful handling for throughput results

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklbench/benchmarks/throughput_worker.py
+++ b/sklbench/benchmarks/throughput_worker.py
@@ -1,5 +1,5 @@
 # ===============================================================================
-# Copyright 2024 Intel Corporation
+# Copyright 2026 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -18,14 +18,13 @@ import argparse
 import inspect
 import json
 import socket
-import sys
 import time
 from typing import Dict, List, Tuple
 
 from ..datasets import load_data
 from ..datasets.transformer import split_and_transform_data
+from ..utils.barrier import recv_until
 from ..utils.bench_case import get_bench_case_value
-from ..utils.common import convert_to_numpy
 from ..utils.config import bench_case_filter
 from ..utils.custom_types import BenchCase
 from ..utils.logger import logger
@@ -38,16 +37,6 @@ from .sklearn_estimator import (
     validate_estimator_params,
 )
 
-
-def barrier_wait(sock: socket.socket, msg_send: bytes, msg_expect_prefix: bytes):
-    """Send a message and block until response from parent."""
-    sock.sendall(msg_send)
-    data = b""
-    while not data.startswith(msg_expect_prefix):
-        chunk = sock.recv(1024)
-        if not chunk:
-            raise ConnectionError("Barrier socket closed unexpectedly")
-        data += chunk
 
 
 def run_measurement_loop(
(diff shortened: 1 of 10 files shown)


sklbench/benchmarks/throughput_worker.py
```diff
@@ -0,0 +1,247 @@
+# ===============================================================================
+# Copyright 2024 Intel Corporation
```
david-cortes-intel:
```suggestion
# Copyright 2026 Intel Corporation
```

sklbench/benchmarks/throughput_worker.py
```diff
@@ -0,0 +1,247 @@
...
+    validate_estimator_params,
+)
+
+
+def barrier_wait(sock: socket.socket, msg_send: bytes, msg_expect_prefix: bytes):
+    """Send a message and block until response from parent."""
+    sock.sendall(msg_send)
+    data = b""
+    while not data.startswith(msg_expect_prefix):
+        chunk = sock.recv(1024)
```
david-cortes-intel: Would be useful to put some comment about the choice of this particular number.

david-cortes-intel:
@avolkov-intel What would be the advantage of using sockets like this instead of just launching multiple processes?

avolkov-intel:
> @avolkov-intel What would be the advantage of using sockets like this instead of just launching multiple processes?

Each process has some preprocessing time, then the task that we want to measure is executed, then there's some postprocessing of results. I want the middle part to be executed simultaniously on each of the different processes to simulate  the full load. Therefore we need to find some way to synchronize and put some kind of barrier. To achieve that we need to either use some sort of MPI barrier (which I don't want to use not to add new deps) or to set up some communication channel between child processes and parent process.

Another thing is that I want the child processes to be launched with numactl restriction on CPUs to ensure that limitation is set on the OS level. This limits the number of ways parent process can communicate with child processes  and using sockets is one of them

david-cortes-intel:
@avolkov-intel Wouldn't you get the same result from launching sub-processes with `numactl` and then making the python block wait until each process finishes before continuing?

PR #214, opened 2026-05-27
Calculate number of nodes in the tree estimators

avolkov-intel:
## Description

- Automatically report total number of tree nodes (`n_nodes`) for Random Forest and Extra Trees estimators after training
- Register `n_nodes` as an indifferent metric in the report so it's compared across libraries rather than treated as a grouping parameter

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/sklbench/benchmarks/sklearn_estimator.py
+++ b/sklbench/benchmarks/sklearn_estimator.py
@@ -134,6 +134,16 @@ def get_subset_metrics_of_estimator(
                 and isinstance(iterations[0], Union[Numeric, NumpyNumeric].__args__)
             ):
                 metrics.update({"iterations": int(iterations[0])})
+        if hasattr(estimator_instance, "estimators_"):
+            estimators_with_trees = [
+                t
+                for t in estimator_instance.estimators_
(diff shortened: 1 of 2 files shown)


sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -134,6 +134,16 @@ def get_subset_metrics_of_estimator(
                 and isinstance(iterations[0], Union[Numeric, NumpyNumeric].__args__)
             ):
                 metrics.update({"iterations": int(iterations[0])})
+        if hasattr(estimator_instance, "estimators_"):
```
david-cortes-intel, line 137: This will also be set in other types of estimators - for example:
https://scikit-learn.org/stable/modules/generated/sklearn.multiclass.OneVsRestClassifier.html

Perhaps it could check if the first entry there is of tree type.

Also what about cuML? I guess it could be done by calling `as_sklearn()`:
https://docs.rapids.ai/api/cuml/nightly/pickling_cuml_models/#Converting-Between-cuML-and-scikit-learn-Models

PR #218, opened 2026-07-10
Update sklearnex gpu to use array_api_dispatch and dpnp

ethanglaser:
## Description

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,6 @@ conda env create -n sklearn -f envs/conda-env-sklearn.yml
 conda env create -n rapids --solver=libmamba -f envs/conda-env-rapids.yml
 ```
 
-GPU benchmarks using `sklearnex`'s `device: gpu` rely on scikit-learn's Array API support (`array_api_dispatch`), which requires `SCIPY_ARRAY_API=1` to be set in the environment before running:
-
-```bash
-export SCIPY_ARRAY_API=1
-```
-
 ## 🚀 How To Use Scikit-learn_bench
 
 ### Benchmarks Runner
--- a/README.md
+++ b/README.md
@@ -40,6 +40,12 @@ conda env create -n sklearn -f envs/conda-env-sklearn.yml
 conda env create -n rapids --solver=libmamba -f envs/conda-env-rapids.yml
 ```
 
+GPU benchmarks using `sklearnex`'s `device: gpu` rely on scikit-learn's Array API support (`array_api_dispatch`), which requires `SCIPY_ARRAY_API=1` to be set in the environment before running:
+
+```bash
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -40,6 +40,12 @@ conda env create -n sklearn -f envs/conda-env-sklearn.yml
 conda env create -n rapids --solver=libmamba -f envs/conda-env-rapids.yml
 ```
 
+GPU benchmarks using `sklearnex`'s `device: gpu` rely on scikit-learn's Array API support (`array_api_dispatch`), which requires `SCIPY_ARRAY_API=1` to be set in the environment before running:
```
david-cortes-intel, line 43: Sounds like it could be set inside the python process instead.

ethanglaser: It _could_ but I don't think this is a great way to handle it

david-cortes-intel: @ethanglaser What would be the downside?

sklbench/benchmarks/sklearn_estimator.py
```diff
@@ -147,6 +147,10 @@ def get_subset_metrics_of_estimator(
             and getattr(estimator_instance, "probability") == False
         ):
             y_pred_proba = convert_to_numpy(estimator_instance.predict_proba(x))
+            # GPU/float32 predict_proba rows may drift outside roc_auc_score's
```
david-cortes-intel, line 150: This should be fixed in sklearnex instead. There is a code branch like this for random forests. If it happens for other classes, needs to be added for them too.

configs/common/sklearn.json
```diff
@@ -6,12 +6,18 @@
...
+        "sklearn-ex[cpu,gpu] implementations": [
+            { "algorithm": { "library": "sklearn", "device": "cpu" } },
+            { "algorithm": { "library": "sklearnex", "device": "cpu" } },
+            {
+                "algorithm": {
+                    "library": "sklearnex",
+                    "device": "gpu",
+                    "sklearn_context": { "array_api_dispatch": true }
+                },
+                "data": { "format": "dpnp" }
```
avolkov-intel, line 18: please double check that C order is used

sklbench/benchmarks/sklearn_estimator.py
ethanglaser: Removed

configs/common/sklearn.json
ethanglaser: explicitly added it

README.md
ethanglaser: updated

david-cortes-intel (approved):

PR #219, opened 2026-08-17
Add combined results and plot generation

avolkov-intel:
### New CLI options
  - `--combined-results` — builds two extra sheets:
    - **All cases** — all algorithm results on one page (group name, sklearn time[ms], sklearnex time[ms], speedup, then parameters). KNN is split into
  brute/kd_tree; rows sorted by dtype then dataset; each group ends with per-dtype and total `GEOMEAN` formula rows with a per-group speedup color scale.
    - **Summary (for plots)** — one row per algorithm referencing the group GEOMEAN cells (total + fp32 + fp64), split into Training/Inference sections with
  section and total geomeans.
  - `--draw-plots` / `--plot-output <path>` — renders a Training/Inference speedup bar chart (geomean per algorithm, combining fp32 & fp64) with
  Intel-branded titles and disclaimer footnote.

### Other changes
  - `--performance-stability-metrics` now additionally emits a `median time[ms]` column (alongside existing `1st run time[ms]`, `time CV`), included in the
  individual sheets and "All cases".

  Use `--compatibility-mode` together with `--combined-results` so sklearn/sklearnex rows merge correctly.

**Testing**

</details>


david-cortes-intel:
@avolkov-intel Please document these options in the .md files where the other options are described.

avolkov-intel:
> @avolkov-intel Please document these options in the .md files where the other options are described.

Done

avolkov-intel:
CI failures are not related to this PR

sklbench/report/compatibility.py
```diff
@@ -34,6 +34,12 @@ def transform_results_to_compatible(results: pd.DataFrame):
                 "min_bin_size",
             ],
         )
+    if (
+        "n_jobs" in results.columns
+        and results["n_jobs"].isna().any()
+        and results["n_jobs"].notna().any()
```
ethanglaser, line 40: Would it make more sense to drop `and results["n_jobs"].notna().any()` and just drop n_jobs col if any are na? Or what is the reason for this change

sklbench/report/implementation.py
```diff
@@ -346,6 +356,281 @@ def apply_rules_for_sheet(sheet, perf_color_scale, quality_color_scale):
...
+
+def write_all_cases_2_sheet(dfs, wb):
+    """
+    Write 'All cases' sheet: one block per algorithm group with columns
+      Algorithm | sklearn time[ms] | sklearnex time[ms] | Speedup | <stability> | <parameters>.
+    KNN groups are split by method (brute/kd_tree); rows are sorted by dtype then dataset.
+    Each block ends with per-dtype and total GEOMEAN formula rows plus a speedup color scale.
+    Returns list of (group_name, total_geomean_row, {dtype: geomean_row}) for the summary sheet.
+    """
+    KNN = ("KNeighborsClassifier", "KNeighborsRegressor")
```
ethanglaser, line 421: Is there a way to make this knn-specific logic more flexible? ie right now we have 2 different knn methods but what it we wanted to use 2 different LogReg solvers - would we need to keep adding in algo-specific logic or can we set it up to be more general?

avolkov-intel: Potentially yes, but this would make the code more complex and we don't need it right now

sklbench/report/compatibility.py
avolkov-intel: If we compare sklearn algorithm with sklearnex counterpart, n_jobs parameter might only be present in sklearnex but not in sklearn. In this case the measurements will not be compared and report & plots would not be generated. So the idea of this change if both not nans and nans are present we can drop this column so that reports are properly generated. If all values are na there's no need to drop the column

avolkov-intel: This feature should only be used for comparison of sklearn vs sklearnex btw

sklbench/report/README.md
```diff
@@ -18,6 +18,10 @@ Raw results are converted into a pandas dataframe and the final report is made b
...
+| `--combined-results`                           |        | False                                               |                                                | [EXPERIMENTAL] Creates `All cases` and `Summary (for plots)` sheets combining time and speedup of all algorithms with per-dtype/total GEOMEAN rows (use with `--compatibility-mode`). |
+| `--draw-plots`                                 |        | False                                               |                                                | [EXPERIMENTAL] Draws Training/Inference speedup bar charts from the combined results (requires `--combined-results`).             |
+| `--plot-output`                                | str    | None                                                |                                                | [EXPERIMENTAL] Output file path for plots (e.g. `plots.png`); if unset, plots are shown interactively.                           |
```
david-cortes-intel, line 24: Does this require passing `--draw-plots`?

avolkov-intel: Well technically no, you can provide it without draw plots it would be just ignored

david-cortes-intel: Please add that kind of detail in these docs.

david-cortes-intel:
@avolkov-intel Would this work when comparing against cuml? If so, how would it handle cases where parameters do not match?

sklbench/report/README.md
```diff
@@ -18,6 +18,10 @@ Raw results are converted into a pandas dataframe and the final report is made b
...
+| `--performance-stability-metrics`</br>`-psm`   |        | False                                               |                                                | Adds performance stability metrics (`1st run time[ms]`, `1st-mean run ratio`, `median time[ms]`, `time CV`) to the report.       |
+| `--combined-results`                           |        | False                                               |                                                | [EXPERIMENTAL] Creates `All cases` and `Summary (for plots)` sheets combining time and speedup of all algorithms with per-dtype/total GEOMEAN rows (use with `--compatibility-mode`). |
+| `--draw-plots`                                 |        | False                                               |                                                | [EXPERIMENTAL] Draws Training/Inference speedup bar charts from the combined results (requires `--combined-results`).             |
```
david-cortes-intel, line 23: This option without `--plot-output` doesn't appear to do anything. It doesn't show anything interactive for me, or at least not when used in an SSH session.

david-cortes-intel:
@avolkov-intel Please add information about how to generate a publication-grade plot, with details in the bottom text filled in.

sklbench/report/implementation.py
```diff
@@ -346,6 +356,281 @@ def apply_rules_for_sheet(sheet, perf_color_scale, quality_color_scale):
...
+    STABILITY = ["1st run time[ms]", "median time[ms]", "time CV"]
+    ws = wb.create_sheet(title="All cases", index=1)
+    row, geomean_rows = 1, []
+
+    def geomean_row(label, r0, r1):
```
david-cortes-intel, line 426: Please make `row` a parameter.

sklbench/report/implementation.py
```diff
@@ -346,6 +356,281 @@ def apply_rules_for_sheet(sheet, perf_color_scale, quality_color_scale):
...
+    # Convert back to multi-index columns if there were multi-index columns
+    if "|" in ordered_cols[0] if ordered_cols else False:
+        df.columns = pd.MultiIndex.from_tuples(
+            [tuple(col.split("|")) for col in df.columns]
+        )
+
+    return df
+
+
+def write_all_cases_2_sheet(dfs, wb):
```
david-cortes-intel, line 413: Please add a description of what `wb` is. Better yet if you can add type hints.

sklbench/report/implementation.py
david-cortes-intel, line 421: What about LocalOutlierFactor and NearestNeighbors?

sklbench/report/implementation.py
```diff
@@ -346,6 +356,281 @@ def apply_rules_for_sheet(sheet, perf_color_scale, quality_color_scale):
...
+            groups = {df_name: df}
+
+        for name, gdf in groups.items():
+            sort_cols = [
+                c
+                for c in (("parameter", "dtype"), ("parameter", "dataset"))
+                if c in gdf.columns
+            ]
+            if sort_cols:
+                gdf = gdf.sort_values(sort_cols, kind="mergesort")
```
david-cortes-intel, line 484: Is this trying to do a stable sort? Otherwise, why pass the parameter?

sklbench/report/implementation.py
```diff
@@ -346,6 +356,281 @@ def apply_rules_for_sheet(sheet, perf_color_scale, quality_color_scale):
...
+                dtype_rows = {}
+                dtype_col = ("parameter", "dtype")
+                if dtype_col in gdf.columns:
+                    dtypes = gdf[dtype_col].astype(str).tolist()
+                    i = 0
+                    while i < len(dtypes):
+                        j = i
+                        while j + 1 < len(dtypes) and dtypes[j + 1] == dtypes[i]:
+                            j += 1
+                        dtype_rows[dtypes[i]] = geomean_row(
```
david-cortes-intel, line 513: Since this one doesn't embed geomean as a formula, maybe the code could be a lot more readable if this were to pre-generate subtables as dataframes and store them in a list to write at the end.

sklbench/report/implementation.py
```diff
@@ -346,6 +356,281 @@ def apply_rules_for_sheet(sheet, perf_color_scale, quality_color_scale):
...
+        "speedup fp64",
+    ]
+
+    def refs_for_row(row_num):
+        if row_num is None:
+            return [None, None, None]
+        return [f"={src_sheet_name}!{c}{row_num}" for c in "BCD"]
+
+    def write_section(title, rows):
+        nonlocal current_row
```
david-cortes-intel, line 587: Please switch it to parameter.

sklbench/report/implementation.py
```diff
@@ -370,6 +655,256 @@ def write_environment_info(results, workbook):
                 new_ws.append([None])
 
 
+def draw_summary_plots(all_cases_df: pd.DataFrame, output_file: str = None):
```
david-cortes-intel, line 658: Would be helpful to have a sub-function to generate the plot returning matplotlib objects (plt, fig, axes).

ethanglaser (approved):
LGTM if it works for you. But address David's comments to his satisfaction before merge.

sklbench/report/implementation.py
avolkov-intel: We currently don't report performance for them, so they are not present

sklbench/report/implementation.py
avolkov-intel: Yes, the idea is to have stable sort

sklbench/report/implementation.py
avolkov-intel: In fact the idea is indeed to embed geomean as a formula. On review they often want formulas to be present in excel spreadsheet rather than calculated outside of it for clarity. You can see here we insert it as a formula: https://github.com/IntelPython/scikit-learn_bench/pull/219/changes#diff-24c57e4b1e3d9bbd1fa3cd538fca5b437e68987456662773afc60a67afdeabc5R627

avolkov-intel:
> @avolkov-intel Would this work when comparing against cuml? If so, how would it handle cases where parameters do not match?

no, currently it's only designed to work for sklearn vs sklearnex comparison

avolkov-intel:
Addressed comments from David

PR #220, opened 2026-08-31
Update Xgboost Benchmarks

avolkov-intel:
## Description 

In this PR the new configs for xgboost benchmarks are introduced. Motivation is to make the use cases closer to real-world usage scenarios and cover some features from the recent xgboost versions.

- Drop some use cases (datasets based on image data, hepmass from regression tasks)
- Add new datasets from this paper: https://arxiv.org/abs/2506.16791
- Add cases with categorical features and missing values to xgboost configs
- Add task_dispatch decorator to enable usage same data with different targets

<details>

<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/configs/common/xgboost.json
+++ b/configs/common/xgboost.json
@@ -6,7 +6,8 @@
                     "device": "cpu",
                     "estimator_params": {
                         "tree_method": "hist",
-                        "n_jobs": "[SPECIAL_VALUE]physical_cpus"
+                        "n_jobs": "[SPECIAL_VALUE]physical_cpus",
+                        "enable_categorical": true
                     },
                     "enable_modelbuilders": false
                 }
@@ -16,7 +17,8 @@
                     "device": "gpu",
                     "estimator_params": {
                         "tree_method": "hist",
-                        "n_jobs": "[SPECIAL_VALUE]physical_cpus"
+                        "n_jobs": "[SPECIAL_VALUE]physical_cpus",
+                        "enable_categorical": true
                     }
                 },
                 "data": { "format": "cudf" }
--- a/sklbench/datasets/common.py
+++ b/sklbench/datasets/common.py
@@ -59,7 +59,12 @@ def get_filenames_by_prefix(directory: str, prefix: str) -> List[str]:
 
 def load_data_file(filepath, extension):
     if extension == "parq":
-        data = pd.read_parquet(filepath)
+        # Read with the same engine used for writing (fastparquet). Reading a
+        # fastparquet-written file with the default "auto"/pyarrow engine drops
+        # pandas "category" dtype (string categories come back as object),
+        # which would prevent XGBoost's native categorical handling from ever
+        # seeing category columns loaded from cache.
+        data = pd.read_parquet(filepath, engine="fastparquet")
     elif extension.endswith("npz"):
         npz_content = np.load(filepath)
         if extension == "npz":
--- a/sklbench/datasets/common.py
+++ b/sklbench/datasets/common.py
@@ -34,19 +34,19 @@ from ..utils.custom_types import Array
 from ..utils.logger import logger
 
 # NB: non-registered data components and extensions will not be found by loader
-KNOWN_DATA_COMPONENTS = ["x", "y"]
+KNOWN_DATA_COMPONENTS = ["x", "y", "y_cls", "y_reg"]
 KNOWN_DATA_EXTENSIONS = ["parq", "npz", "csr.npz"]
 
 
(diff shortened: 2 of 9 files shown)


configs/common/xgboost.json
```diff
@@ -6,7 +6,8 @@
                     "device": "cpu",
                     "estimator_params": {
                         "tree_method": "hist",
-                        "n_jobs": "[SPECIAL_VALUE]physical_cpus"
+                        "n_jobs": "[SPECIAL_VALUE]physical_cpus",
+                        "enable_categorical": true
```
david-cortes-intel, line 10: This is enabled by default in the latest versions. Which xgboost version are you using where it makes an effect?

sklbench/datasets/common.py
```diff
@@ -59,7 +59,12 @@ def get_filenames_by_prefix(directory: str, prefix: str) -> List[str]:
 
 def load_data_file(filepath, extension):
     if extension == "parq":
-        data = pd.read_parquet(filepath)
+        # Read with the same engine used for writing (fastparquet). Reading a
+        # fastparquet-written file with the default "auto"/pyarrow engine drops
+        # pandas "category" dtype (string categories come back as object),
+        # which would prevent XGBoost's native categorical handling from ever
+        # seeing category columns loaded from cache.
+        data = pd.read_parquet(filepath, engine="fastparquet")
```
david-cortes-intel, line 67: Wouldn't a better fix be to switch both to pyarrow, or to use polars to read it?

sklbench/datasets/loaders.py
```diff
@@ -476,6 +476,132 @@ def load_fraud(
...
+    predict the propensity of customers to buy new products or services
+    ("appetency"). It is a hard, real-world marketing dataset whose key
+    characteristics are a large number of missing values (every instance
+    has at least one) and heavily unbalanced classes (the positive class
+    is roughly 1.8% of samples). Feature names and categorical values are
+    anonymized and carry no semantic meaning.
+
+    Source: OpenML dataset id 46939 (kddcup09_appetency, small training set).
+
+    Classification task. n_classes = 2.
```
david-cortes-intel, line 500: These comments are redundant given the rest of the code below.

sklbench/datasets/transformer.py
```diff
@@ -26,11 +26,31 @@
...
+    # before the numpy round-trip below flattens the DataFrame and drops this
+    # metadata. The categoricals are restored after converting back to pandas,
+    # so estimators with native categorical support (e.g. XGBoost with
+    # enable_categorical=True) still see "category" columns.
+    categorical_dtypes = None
+    if (
+        dformat == "pandas"
+        and isinstance(data, pd.DataFrame)
+        and any(str(t) == "category" for t in data.dtypes)
+    ):
```
david-cortes-intel, line 42: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.select_dtypes.html

sklbench/datasets/transformer.py
```diff
@@ -44,15 +64,20 @@ def convert_data(
...
+        data = pd.DataFrame(data)
+        if categorical_dtypes is not None:
+            # restore original column names, re-apply the saved CategoricalDtype
+            # to categorical columns, and cast the rest to the requested dtype.
+            data.columns = column_names
+            for column in data.columns:
+                if column in categorical_dtypes:
+                    data[column] = data[column].astype(categorical_dtypes[column])
+                elif numeric_dtype is not None:
+                    data[column] = data[column].astype(numeric_dtype)
```
david-cortes-intel: This can be done without a loop.

configs/common/xgboost.json
avolkov-intel: We need to specify it to ensure comparison with previous xgboost version is correct

sklbench/datasets/common.py
avolkov-intel: If the changes from the current PR are applied benchmark caches are not invalidated, we only change the code that is responsible for reading. If we switch to another engine then existing caches would be invalidated that might silently affect some CI jobs etc. Probably it worth doing later, I just don't want to deal with these changes in the current PR

sklbench/datasets/loaders.py
avolkov-intel: I would say that they are unnecessary, this brief description is really useful when you look at the configs and benchmarking results

david-cortes-intel: But you have the exact same info in a more succinct form right below it.

